PR #82, opened 2015-12-01
Prevent rendering a non-scrollable table when dataset length is 0

jfurrow:
This fixes a bug revealed in the groups table in DCOS UI.

### DIFF
--- a/src/Table/Table.js
+++ b/src/Table/Table.js
@@ -327,7 +327,13 @@ export default class Table extends React.Component {
 
     let style = {};
 
-    if (scrollContainer != null) {
+    if (data.length === 0) {
+      innerContent = (
+        <tbody>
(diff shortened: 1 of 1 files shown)


src/Table/Table.js
```diff
@@ -327,7 +327,13 @@ export default class Table extends React.Component {
 
     let style = {};
 
-    if (scrollContainer != null) {
+    if (data.length === 0) {
+      innerContent = (
+        <tbody>
+          {this.getEmptyRowCell(columns)}
+        </tbody>
+      );
```
rcorral: How does this fix things?

rcorral: nvm I get it.

rcorral:
:ship:

PR #85, opened 2015-12-01
Fix error when rendering empty dataset

jfurrow:
So our previous solution forces the scrollTable to render under certain conditions, but didn't account for an empty dataset. This fixes that.

### DIFF
--- a/src/Table/Table.js
+++ b/src/Table/Table.js
@@ -315,7 +315,7 @@ export default class Table extends React.Component {
     let buildRowOptions = this.props.buildRowOptions;
     let childToMeasure;
 
-    if (itemHeight === 0) {
+    if (itemHeight === 0 && data.length) {
       childToMeasure = this.getRowCells(columns, sortBy, buildRowOptions, idAttribute, data[0]);
     }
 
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @rcorral

src/Table/Table.js
```diff
@@ -315,7 +315,7 @@ export default class Table extends React.Component {
     let buildRowOptions = this.props.buildRowOptions;
     let childToMeasure;
 
-    if (itemHeight === 0) {
+    if (itemHeight === 0 && data.length) {
```
rcorral: Can you describe why you're making this change?

jfurrow: Before we made changes to this yesterday, the table component would render the regular table (as opposed to the `scrollTable`) if the length of the dataset were 0. We changed that because toggling between the `scrollTable` and the regular table prevented gemini from maintaining a reference to its container, which it needs to calculate its height. This worked, but `scrollTable` expects to measure the height of its first child (which is calculated by passing `data[0]` to the measuring function), and when the dataset is 0, there is no first item.

rcorral: cool thanks.

PR #87, opened 2015-12-02
Adding null check before accessing properties

mlunoe:
From stack trace:

### DIFF
--- a/src/Dropdown/Dropdown.js
+++ b/src/Dropdown/Dropdown.js
@@ -34,8 +34,12 @@ export default class Dropdown extends Util.mixin(BindMixin) {
     let obj = Util.find(items, function (item) {
       return item.id === id;
     });
-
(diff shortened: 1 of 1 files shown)


src/Dropdown/Dropdown.js
```diff
@@ -34,8 +34,12 @@ export default class Dropdown extends Util.mixin(BindMixin) {
...
       return item.id === id;
     });
-
-    return obj.selectedHtml || obj.html;
+    
+    if (obj != null) {
+      return obj.selectedHtml || obj.html;
+    }
+    
+    return null;
```
rcorral: Can we write a test for this?

rcorral: ping @mlunoe

kennyt:
konflicts

PR #88, opened 2015-12-02
Add option to turn off scrolling table

jfurrow:


mesosphere-web:
Reviewer: @rcorral

kennyt:
got some tests failing :(

jfurrow:
Weird, I removed that test yesterday...

jfurrow:
Nevermind, it was a different subjective test with the same name.

src/Table/Table.js
```diff
@@ -383,7 +383,8 @@ export default class Table extends React.Component {
 
     // Use scroll table on first render to check if we need to scroll
     // and if content is bigger than its container
-    if (props.useFlex || itemHeight === 0 || itemListHeight > containerHeight) {
+    if ((props.useFlex || itemHeight === 0 || itemListHeight > containerHeight)
+      && props.useScrollTable !== false) {
```
mlunoe: Can't we just default this to `false`, and then check if it is `true`?

mlunoe:
Other than that I think it looks good

src/Table/Table.js
kennyt: dont we want to do `|| props.useScrollTable` ? because we always want to use the regular scroll table no matter what if this is true

kennyt: woops, my bad. scrollTable is the infinite table. ignore that

kennyt:
yeah we should default `useScrollTable` to true and then set it to false whenever we don't want it

jfurrow:
It makes sense to me to default to `true`, I'll do that

src/Table/Table.js
jfurrow: So now it defaults to `true` (because we want to maintain the original behavior, where by default it will determine whether or not it needs to render the scrollTable).

jfurrow:
@kennyt Done!

src/Table/Table.js
```diff
@@ -383,7 +383,8 @@ export default class Table extends React.Component {
 
     // Use scroll table on first render to check if we need to scroll
     // and if content is bigger than its container
-    if (props.useFlex || itemHeight === 0 || itemListHeight > containerHeight) {
+    if ((props.useFlex || itemHeight === 0 || itemListHeight > containerHeight)
+      && props.useScrollTable === true) {
```
kennyt: can we just do `&& props.useScrollTable` without the === true? after that feel free to merge

jfurrow: Done!

PR #90, opened 2015-12-02
Allow dropdown to appear above dropdown button if viewport is too small

jfurrow:
Reading the component's code might be a little confusing, so here's an overview of what happens:
1. The first time the dropdown opens, it renders inside a hidden element with class `dropdown-menu-concealer`
2. Immediately afterwards, `componentDidUpdate` measures the concealed menu, as well as the dropdown button's position in the viewport, and the viewport height, and updates the state with the correct position for the menu (`up` or `down`)
3. The component renders again with the correct `up` or `down` classes & animations, and without the concealer
4. Every subsequent time the dropdown opens, it calculates the direction at the same time it toggles its `open` state.

### DIFF
--- a/src/Dropdown/Dropdown.js
+++ b/src/Dropdown/Dropdown.js
@@ -31,14 +31,13 @@ export default class Dropdown extends Util.mixin(BindMixin) {
     // it's rendered. It's rendered inside a concealed container, so it's okay
     // if it renders in the wrong direction.
     if (this.state.menuHeight == null) {
-      console.log('calculate menu height');
       let dropdownMenuConcealer = this.refs.dropdownMenuConcealer.getDOMNode();
       let menuDirection = this.state.menuDirection;
       let menuHeight = this.state.menuHeight;
 
       if (dropdownMenuConcealer != null) {
         // Get the height and direction of the concealed menu.
-        menuHeight = dropdownMenuConcealer.firstChild.clientHeight;
+        menuHeight = dropdownMenuConcealer.firstChild.clientHeight || 0;
         menuDirection = this.getMenuDirection(menuHeight);
       }
 
@@ -48,8 +47,6 @@ export default class Dropdown extends Util.mixin(BindMixin) {
         menuDirection,
         menuHeight
       });
-    } else {
-      console.log('do not calculate menu height');
     }
   }
 
@@ -170,7 +167,7 @@ export default class Dropdown extends Util.mixin(BindMixin) {
     // the DOM element around while we are measuring it.
     let props = this.props;
     let state = this.state;
-    let dropdownKey = state.menuHeight || 'initial-render';
+    let dropdownKey = 'initial-render';
     let dropdownMenu = null;
     let dropdownMenuClassSet = classNames(
       state.menuDirection,
@@ -187,6 +184,10 @@ export default class Dropdown extends Util.mixin(BindMixin) {
       dropdownStateClassSet,
       props.wrapperClassName
     );
+
+    if (state.menuHeight != null) {
+      dropdownKey = state.menuHeight;
+    }
 
     if (state.isOpen) {
       dropdownMenu = (
--- a/src/Dropdown/Dropdown.js
+++ b/src/Dropdown/Dropdown.js
@@ -39,7 +39,6 @@ export default class Dropdown extends Util.mixin(BindMixin) {
       );
       // Get the height of the concealed menu.
       menuHeight = dropdownMenuConcealer.children[0].clientHeight;
-      // Get the
       menuDirection = this.getMenuDirection(menuHeight);
 
       // Only set state if the properties actually changed.
--- a/src/Dropdown/Dropdown.js
+++ b/src/Dropdown/Dropdown.js
@@ -169,30 +169,32 @@ export default class Dropdown extends Util.mixin(BindMixin) {
   render() {
     // Set a key based on the menu height so that React knows to keep the
     // the DOM element around while we are measuring it.
-    let dropdownKey = this.state.knowMenuHeight || 'initial-render';
+    let props = this.props;
+    let state = this.state;
+    let dropdownKey = state.knowMenuHeight || 'initial-render';
     let dropdownMenu = null;
     let dropdownMenuClassSet = classNames(
-      this.state.menuDirection,
-      this.props.dropdownMenuClassName
+      state.menuDirection,
+      props.dropdownMenuClassName
     );
     let dropdownStateClassSet = {
-      'open': this.state.isOpen
+      'open': state.isOpen
     };
-    let items = this.props.items;
+    let items = props.items;
     let transitionName =
-      `${this.props.transitionName}-${this.state.menuDirection}`;
+      `${props.transitionName}-${state.menuDirection}`;
     let wrapperClassSet = classNames(
-      this.state.menuDirection,
+      state.menuDirection,
       dropdownStateClassSet,
-      this.props.wrapperClassName
+      props.wrapperClassName
     );
 
-    if (this.state.isOpen) {
+    if (state.isOpen) {
       dropdownMenu = (
         <span className={dropdownMenuClassSet}
           role="menu" ref="dropdownMenu" key={dropdownKey}>
-          <ul className={this.props.dropdownMenuListClassName}>
-            {this.getMenuItems(this.props.items)}
+          <ul className={props.dropdownMenuListClassName}>
+            {this.getMenuItems(props.items)}
           </ul>
         </span>
       );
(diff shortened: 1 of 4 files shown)


mesosphere-web:
Reviewer: @kennyt

src/Dropdown/Dropdown.js
```diff
@@ -19,23 +19,68 @@ export default class Dropdown extends Util.mixin(BindMixin) {
...
+
+    // We need to calculate the position & height of the menu when the user
+    // opens it.
+    if (this.state.isOpen) {
+      let dropdownMenuConcealer = React.findDOMNode(
+        this.refs.dropdownMenuConcealer
+      );
+      // Get the height of the concealed menu.
+      menuHeight = dropdownMenuConcealer.children[0].clientHeight;
+      // Get the
```
kennyt: this comment might be incomplete, but maybe i'm wrong

jfurrow: hahaha, fixed!

src/Dropdown/Dropdown.js
```diff
@@ -97,36 +156,66 @@ export default class Dropdown extends Util.mixin(BindMixin) {
...
     this.setState({
+      menuDirection,
       isOpen: !this.state.isOpen
     });
   }
 
   render() {
+    // Set a key based on the menu direction so that React knows to keep the
+    // the menu around while we are measuring it.
+    let dropdownKey = this.state.knowMenuHeight || "initial-render";
```
kennyt: Gotta use single-quotes

jfurrow: Woops, done!

src/Dropdown/Dropdown.js
```diff
@@ -63,6 +100,28 @@ export default class Dropdown extends Util.mixin(BindMixin) {
...
+  getSpaceAroundDropdown() {
+    let dropdownWrapper = React.findDOMNode(this.refs.dropdownWrapper);
+    let position = dropdownWrapper.getBoundingClientRect();
+    let viewportHeight = global.window.innerHeight;
+
+    // Calculate the distance from the top of the viewport to the top of the
+    // dropdown button as well as the distance from the bottom of the viewport
+    // to the bottom of the dropdown button.
+    return {
+      top: position.top,
```
mlunoe: Do we actually use this?

jfurrow: Nope! Good call, I forgot that we're just testing the distance from the bottom of the viewport.

jfurrow: Removed!

src/Dropdown/Dropdown.js
```diff
@@ -97,36 +156,66 @@ export default class Dropdown extends Util.mixin(BindMixin) {
...
+    );
     let dropdownStateClassSet = {
       'open': this.state.isOpen
     };
-    let dropdownMenu = null;
     let items = this.props.items;
+    let transitionName =
+      `${this.props.transitionName}-${this.state.menuDirection}`;
     let wrapperClassSet = classNames(
+      this.state.menuDirection,
```
mlunoe: #nit can we get `state` and `props` as variables, so we don't access them multiple times?

src/Dropdown/Dropdown.js
```diff
@@ -97,36 +156,66 @@ export default class Dropdown extends Util.mixin(BindMixin) {
...
-            {this.getMenuItems(items)}
+            {this.getMenuItems(this.props.items)}
           </ul>
         </span>
       );
     }
 
+    // If we don't know the menu's height, we render it invisibly and then
+    // immediately measure its height in #componentDidUpdate, which will change
+    // the sate and trigger another render.
```
mlunoe: correction: "state"

src/Dropdown/Dropdown.js

src/Dropdown/Dropdown.js

mlunoe:
I think it looks good @jfurrow!
I think we should give it a :shipit:. However, we should probably think about handling this problem, where the dropdown/dropup is not shown outside of its container in the future - but not now:

jfurrow:
Cool! So that's because the container has `overflow: hidden`, which we can't break out of

src/Dropdown/Dropdown.js
```diff
@@ -19,23 +19,60 @@ export default class Dropdown extends Util.mixin(BindMixin) {
   constructor() {
     super();
     this.state = {
+      menuDirection: 'down',
+      menuHeight: null,
       isOpen: false,
-      selectedID: null
+      selectedID: null,
+      knowMenuHeight: false
```
rcorral: Do we need this property?
Wouldn't it be good enough to check if `menuHeight` is not `null`?

jfurrow: Yeah, initially that's what I checked for. I added the separate `knowMenuHeight` variable because we may want to recalculate the height of the menu if the items inside of it are dynamic. But we're not doing that now — would you prefer to just remove the `knowMenuHeight` prop?

src/Dropdown/Dropdown.js
```diff
@@ -97,36 +152,68 @@ export default class Dropdown extends Util.mixin(BindMixin) {
...
           </ul>
         </span>
       );
     }
 
-    if (this.props.transition) {
+    // If we don't know the menu's height, we render it invisibly and then
+    // immediately measure its height in #componentDidUpdate, which will change
+    // the state and trigger another render.
+    if (state.isOpen && state.knowMenuHeight === false) {
```
rcorral: Could we join this `if` with the one above so that if the dropdown is `open` but we don't know the height we don't end up going into two `if` statements.

src/Dropdown/Dropdown.js
```diff
@@ -19,23 +19,60 @@ export default class Dropdown extends Util.mixin(BindMixin) {
...
+    // it's rendered. It's rendered inside a concealed container, so it's okay
+    // if it renders in the wrong direction.
+    if (!this.state.knowMenuHeight) {
+      let dropdownMenuConcealer = React.findDOMNode(
+        this.refs.dropdownMenuConcealer
+      );
+      let menuDirection = this.state.menuDirection;
+      let menuHeight = this.state.menuHeight;
+
+      if (dropdownMenuConcealer != null) {
```
rcorral: Would we ever not go into this `if`? is it necessary?

src/Dropdown/Dropdown.js
```diff
@@ -19,23 +19,60 @@ export default class Dropdown extends Util.mixin(BindMixin) {
...
+        // Get the height and direction of the concealed menu.
+        menuHeight = dropdownMenuConcealer.children[0].clientHeight;
```
rcorral: Could we find the `child` in a different way? 
You're accessing reacts internal API. If we can do this with like `querySelector` that would be nicer.

mlunoe:
I was thinking a solution with `position: absolute` or something, because this probably should never happen. But again, this is not for this PR.

src/Dropdown/Dropdown.js
rcorral: I see, so if we re-render the dropdown while it's already open it'll be smart enough to toggle where it's placed? Do you think that's the best UX?

jfurrow: No, it won't toggle where it's placed if the dropdown is already open and the items change. I meant we could tell it to recalculate its height before it's opened by setting  `knowMenuHeight` to `false`, and we could do this in `componentWillReceiveProps` if its number of items change.

src/Dropdown/Dropdown.js
jfurrow: Yes, it's necessary for the tests to pass.

jfurrow:
The dropdown menu is already `position: absolute`, but the code wrapper has `overflow: hidden` which we can't get out of.

src/Dropdown/Dropdown.js
jfurrow: I can do this, but we're going to end up repeating ourselves, so we have to keep the dropdown menu updated in two locations. Here's what I think you're suggesting:
```javascript
if (state.isOpen) {
  if (state.knowMenuHeight) {
    dropdownMenu = (
      <span className={dropdownMenuClassSet}
        role="menu" ref="dropdownMenu" key={dropdownKey}>
        <ul className={props.dropdownMenuListClassName}>
          {this.getMenuItems(props.items)}
        </ul>
      </span>
    );
  } else {
    dropdownMenu = (
      <div className="dropdown-menu-concealer" ref="dropdownMenuConcealer">
        <span className={dropdownMenuClassSet}
          role="menu" ref="dropdownMenu" key={dropdownKey}>
          <ul className={props.dropdownMenuListClassName}>
            {this.getMenuItems(props.items)}
          </ul>
        </span>
      </div>
    );
  }
}
```

src/Dropdown/Dropdown.js

src/Dropdown/Dropdown.js
jfurrow: Removed!

rcorral:
:shipit:

PR #91, opened 2015-12-03
Add Confirm Modal

rcorral:


jfurrow:
This looks good to me. It might be nice to specify the number of buttons in the future, but I think this works well for now!

jfurrow:
Also, I hate the padding on the modal footer but it looks like that will be overridden in DCOS UI.

PR #105, opened 2015-12-06
fix outdated eslint rule

olegakbarov:
Are you guys interested in more strict `.eslintrc`? I could work on it.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -39,7 +39,7 @@
     "space-before-function-paren": [2, {
       "anonymous": "always", "named": "never"
     }],
-    "spaced-line-comment": [2, "always"],
+    "spaced-comment": [2, "always"],
     // "max-len": [2, 80, 4], TODO: DCOS-914
     "valid-jsdoc": [2, {
       "requireReturn": false,


mesosphere-web:
Reviewer: @jfurrow

rcorral:
@olegakbarov thanks for your contribution! Would love to have you help us out with a more strict eslint file.

.eslintrc
```diff
@@ -39,7 +39,7 @@
     "space-before-function-paren": [2, {
       "anonymous": "always", "named": "never"
     }],
-    "spaced-line-comment": [2, "always"],
+    "spaced-comment": [2, "always"],
```
rcorral: I'm not sure this will work since we're not using the 1.x version of eslint.

olegakbarov: Yea, i got to check this. Are you're interested in migration towards >=1.0.0 ?

rcorral: That sounds good, I think it makes sense!
I would just want to keep parity at the very minimum with the rules that we currently employ.

olegakbarov: I took all of your rules and only added a few new. You can [take a look at the result](https://gist.github.com/olegakbarov/585bdf5232f181639487). What do you think?

rcorral:
We've upgraded to Babel6

PR #108, opened 2015-12-09
Updating dropdown to not have state

mlunoe:
Loosing the state from Dropdown makes it more flexible, so we always tell it what to display, and it does not make assumptions to what should be displayed

### DIFF
--- a/docs/src/Dropdown/index.js
+++ b/docs/src/Dropdown/index.js
@@ -9,11 +9,13 @@ class DropdownExample extends React.Component {
 
     this.state = {
       selectedIDOne: 'foo',
-      selectedIDTwo: 'foo'
+      selectedIDTwo: 'foo',
+      selectedIDThree: 'foo'
     };
 
     this.onItemSelectionOne = this.onItemSelectionOne.bind(this);
     this.onItemSelectionTwo = this.onItemSelectionTwo.bind(this);
+    this.onItemSelectionThree = this.onItemSelectionThree.bind(this);
   }
 
   onItemSelectionOne(item) {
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @jfurrow

docs/src/Dropdown/index.js
```diff
@@ -6,13 +6,22 @@ class DropdownExample extends React.Component {
 
   constructor() {
     super();
-    this.dropdownSelected = 'foo';
-    this.onItemSelection = this.onItemSelection.bind(this);
+
+    this.state = {
+      selectedIDOne: 'foo',
```
mlunoe: Making this explicit, so the examples can follow them completely

jfurrow:
This will require some refactoring everywhere that currently implements the dropdown, correct?

rcorral:
Yeah ^
So this is my proposal to this problem.
* We rename the current `selectedID` property to `initialID`
* We add a new property called `forceSelectedID`, which when set, it won't allow the dropdown to set its own state on `handleItemClick` but it still calls the `this.props.onItemSelection` callback.

And I think that's it, pretty simple. This doesn't cause a ripple effect of breaking changes and us having to keep state on the component that implements the dropdown.

What do you guys think?

jfurrow:
I think I would prefer that solution, as it still keeps the component tidy and easy to implement.

mlunoe:
I do see your points, but I do not mind the refactoring and I think this will keep our code simple and cleaner if this pattern is used for other components. It feels like adding more code to handle edge cases.
However, you are two against one, so I can go with your solution.
I'm closing this until I have the change up.

mlunoe:
Continued here: PR https://github.com/mesosphere/reactjs-components/pull/116

PR #118, opened 2015-12-15
Putting `modal-container` back into transitions

mlunoe:


mesosphere-web:
Reviewer: @rcorral

src/Modal/ModalContents.js
```diff
@@ -249,13 +251,9 @@ export default class ModalContents extends Util.mixin(BindMixin) {
...
   render() {
     let props = this.props;
-    let containerClass = {};
-    if (props.open) {
-      containerClass[props.containerClass] = true;
-    }
 
     return (
-      <div className={classNames(containerClass)}>
+      <div>
```
rcorral: I think we need a class at this level. `modalWrapperClass`.

mlunoe: Any particular reason?

rcorral: So you can easily target the modal.

jfurrow: Sure, I say let's add the optional class just in case. However I don't think it will be super useful because the element doesn't animate, and it will always be in the DOM, regardless of whether the modal is open or closed.

mlunoe: ok, done

jfurrow:
:ship:

src/Modal/ModalContents.js
```diff
@@ -344,5 +342,6 @@ ModalContents.propTypes = {
   headerContainerClass: PropTypes.string,
   innerBodyClass: PropTypes.string,
   modalClass: PropTypes.string,
+  modalWrapperClass: PropTypes.string,
```
rcorral: Needs to be added to docs.

rcorral:
:ship

PR #123, opened 2015-12-16
Better solution to forceSelectedID

mlunoe:


mesosphere-web:
Reviewer: @rcorral

src/Dropdown/Dropdown.js
```diff
@@ -290,6 +280,8 @@ export default class Dropdown extends Util.mixin(BindMixin) {
       );
     }
 
+    let selectedID = props.forceSelectedID || state.selectedID;
```
kennyt: maybe we can abstract this logic out into a method called `getSelectedID`

jfurrow:
I like this, but I'm not crazy about the prop `forceSelectedID`. I think we should use the word "persistent," like `persistentID` or `persistentItemID` perhaps?

src/Dropdown/Dropdown.js
rcorral: +1

rcorral:
I like the name @jfurrow came up with!

src/Dropdown/Dropdown.js
mlunoe: Good point!

mlunoe: done

mlunoe:
Changed name!

PR #126, opened 2015-12-18
Modal adapts to content changing height

kennyt:


mesosphere-web:
Reviewer: @daiyiwhale

src/Modal/ModalContents.js
```diff
@@ -63,8 +64,22 @@ export default class ModalContents extends Util.mixin(BindMixin) {
   }
 
   checkHeight() {
+    let prevContentHeight = this.heightInfo.innerContentHeight;
```
mlunoe: Could we get some tests in for this function?
I wanna make sure that we are doing it right with these extra calculations

mlunoe:
I cannot come up with a more elegant solution right now, and I am guessing this is an intermediate solution, until we can spend some more time on it?

I really hoped we wouldn't make this component more complex than it is

kennyt:
yeah, I think we should do a height calculation overhaul later on. but surprisingly, the modal has been reliable (other than the content height change, which this should fix)

src/Modal/ModalContents.js
kennyt: alright, made some tests

PR #129, opened 2015-12-30
Upgrade dependencies

rcorral:
Upgraded:
* canvas-ui
* classnames

Removed
* body-parser

Other
* Moved react and react-gemini to devDependencies and to peerDependencies.


mesosphere-web:
Reviewer: @daiyiwhale

package.json
```diff
@@ -19,12 +19,9 @@
   ],
   "main": "main.js",
   "dependencies": {
-    "body-parser": "^1.13.2",
```
rcorral: This wasn't being used.

src/overrides/gemini-scrollbar.less
```diff
@@ -1,4 +1,4 @@
-@import (inline) "../../node_modules/react-gemini-scrollbar/node_modules/gemini-scrollbar/gemini-scrollbar.css";
+@import (inline) "../../../../node_modules/react-gemini-scrollbar/node_modules/gemini-scrollbar/gemini-scrollbar.css";
```
kennyt: this fails when I try to run `livereload` and see the docs

rcorral: Fixed, now people using reactjs components will be required to add it themselves.

src/overrides/gemini-scrollbar.less
```diff
@@ -1,4 +1,5 @@
-@import (inline) "../../node_modules/react-gemini-scrollbar/node_modules/gemini-scrollbar/gemini-scrollbar.css";
+// Make sure to import the following path before including this file
+// @import (inline) "node_modules/react-gemini-scrollbar/node_modules/gemini-scrollbar/gemini-scrollbar.css";
```
kennyt: so the end user of `reactjs-components` would have to import the `gemini-scrollbar.css` correctly?

daiyiwhale:
lgtm aside from Kenny's comment!

src/overrides/gemini-scrollbar.less
rcorral: Correct. I added some docs to the readme and the kitchensink.
This really is the only way to do it that allows us to not install react because react is a peerDependency for react-gemini-scrollbar.

kennyt: is there any way to do it so that the developers would have to do it, rather than the end user? it's just more overhead for people wanting to use this module right? (after forcing them to install gemini in the first place)

README.md
```diff
@@ -30,18 +30,24 @@ A library of reusable React components. For examples, take a look at our
...
+        // Content
+      <Modal />
+    );
+  }
   ```
 
+4. Import CSS files – reactjs-components uses a custom scrollbar.
+
+  ```less
+  @import (inline) "node_modules/react-gemini-scrollbar/node_modules/gemini-scrollbar/gemini-scrollbar.css"
```
mlunoe: Shouldn't this be `"../../../node_modules/[...]"`?

src/overrides/gemini-scrollbar.less
rcorral: Check out my last commit.

PR #130, opened 2015-12-30
Do not cap height of Table

kennyt:
# Changes
* removes GeminiScrollbar from Table
* does not cap height of Table
* adds containerSelector functionality (ex. pass in `.container` and it will find the closest ancestor and use that as the container)
* only use a single table, instead of having two tables (for scrollbar)

### DIFF
--- a/src/Table/Table.js
+++ b/src/Table/Table.js
@@ -1,6 +1,5 @@
 import classNames from 'classnames';
 import DOMUtil from '../Util/DOMUtil';
-import GeminiScrollbar from 'react-gemini-scrollbar';
 import React, {PropTypes} from 'react/addons';
 import Util from '../Util/Util';
 import VirtualList from '../VirtualList/VirtualList';
@@ -83,14 +82,6 @@ export default class Table extends React.Component {
     let state = this.state;
     let refs = this.refs;
 
-    if (state.scrollContainer == null && refs.gemini != null) {
-      // Get the Gemini scroll view as the container view.
-      // We need this since both Gemini and VirtualList need the same element
-      newState.scrollContainer = React.findDOMNode(
-        refs.gemini.refs['scroll-view']
-      );
-    }
-
     if (props.itemHeight == null &&
       state.itemHeight == null &&
       refs.itemHeightContainer != null) {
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/Table/Table.js
```diff
@@ -123,27 +121,10 @@ export default class Table extends React.Component {
       newState.viewportHeight =
         props.windowRatio * DOMUtil.getViewportHeight() - headerHeight;
       // Calculated scroll container height
-      let scrollContainerHeight =
```
kennyt: Don't need scroll container height anymore, it's going to just be as long as it needs to be

src/Table/__tests__/Table-test.js
```diff
@@ -44,17 +44,6 @@ describe('Table', function () {
       .toEqual(4);
   });
 
-  it('should render the proper number of rows', function () {
```
kennyt: We don't have `getRows` anymore, that was only used for when we differentiated between scrollbar/no scrollbar. We always use `VirtualList` now

src/Table/Table.js
```diff
@@ -474,18 +368,5 @@ Table.propTypes = {
   sortBy: PropTypes.shape({
     order: PropTypes.oneOf(['asc', 'desc']),
     prop: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
-  }),
-
-  // Optional property to add transitions or turn them off. Default is off.
-  // Only available for tables that does not scroll
-  transition: PropTypes.bool,
```
kennyt: All of these props are useless now that we don't use `Gemini`. Transition does not work with `VirtualList` yet

src/Table/Table.js
```diff
@@ -383,26 +290,13 @@ export default class Table extends React.Component {
...
-    }
-
-    let styles = {};
-    let className = '';
-
-    if (props.useFlex) {
-      className = 'no-overflow';
-      styles.flexGrow = 1;
-    }
+    let tableContent =
```
kennyt: Always render `VirtualList` instead of doing mass logic in order to determine which table to use

src/Table/Table.js
```diff
@@ -100,50 +96,12 @@ export default class Table extends React.Component {
       ).height;
     }
 
-    if (props.useFlex) {
```
kennyt: This looks extremely intimidating. But all of this logic was to determine the table's height, etc. Since the table just flows as long as it needs to, we can safely remove :+1:

rcorral:
:eyes::ok_hand::eyes::ok_hand::eyes::ok_hand::eyes::ok_hand::eyes: good shit good sHit :ok_hand: that's :heavy_check_mark: some good :ok_hand: :ok_hand: shit right :ok_hand: :ok_hand: there :ok_hand: :ok_hand: :ok_hand: right:heavy_check_mark:there:heavy_check_mark::heavy_check_mark: if I do say so myself :100: :100: i say so :100: that's what I'm talking about right there (chorus: "right there") mMMMMM :100: :ok_hand: :ok_hand: HOOOoOOOooOOooooooo :ok_hand: :ok_hand: :ok_hand: :ok_hand: :ok_hand: :ok_hand: :ok_hand: :100: :ok_hand: :eyes: :eyes: :eyes: :ok_hand: :ok_hand: :ok_hand: Good shit

kennyt:
@rcorral new years celebration is tomorrow. Got started early?

PR #131, opened 2015-12-30
Add emptyMessage prop to show when there is no data in Table

kennyt:


mesosphere-web:
Reviewer: @rcorral

rcorral:
Can we add a test or two?

kennyt:
alright, done

src/Table/__tests__/Table-test.js
```diff
@@ -49,4 +49,32 @@ describe('Table', function () {
...
+          columns={MockTable.columns}
+          data={[]}
+          emptyMessage={this.emptyMessage}
+          idAttribute={this.idAttribute}
+          sortBy={this.sortBy}
+          onSortCallback={this.callback} />
+      );
+    });
+
+    it('should display the custom empty message', function () {
```
rcorral: Can we also have a test case for the inverse? To make sure that `'No data'` is used?

kennyt: alright, done

PR #132, opened 2015-12-30
Pass columns into Table column #getClassName

kennyt:
An example use case for this is if you add class names based on how many columns there are. (Perhaps you wouldn't want to highlight the column if it is the only column)


mesosphere-web:
Reviewer: @jfurrow

rcorral:
:shipit:

PR #138, opened 2016-01-04
Cache row cells instead of rows

kennyt:
Also: added `dontCache` functionality to column props

### DIFF
--- a/src/Table/Table.js
+++ b/src/Table/Table.js
@@ -108,8 +108,8 @@ export default class Table extends React.Component {
     }
   }
 
-  buildUniqueID(cellClassName, cellValue, column) {
-    return column.prop + cellClassName + cellValue;
+  buildUniqueID(cellClassName, cellValue, prop) {
+    return prop + cellClassName + cellValue;
   }
 
   getHeaders(headers, sortBy) {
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

src/Table/Table.js
```diff
@@ -70,9 +73,19 @@ export default class Table extends React.Component {
...
-  componentWillUpdate() {
-    // Clear cached rows
-    this.cachedRows = {};
+  checkFullCache() {
+    if (this.cachedIDs.length < MAX_CACHE_SIZE) {
+      return;
+    }
+
+    let lastID = this.cachedIDs.shift();
+    delete this.cachedCells[lastID];
```
mlunoe: Nice!

src/Table/Table.js
```diff
@@ -167,15 +184,18 @@ export default class Table extends React.Component {
...
-    // Skip build row if we have it memoized
-    if (this.cachedRows[id] == null) {
-      let rowCells = columns.map(function (column, index) {
-        // For each column in the data, output a cell in each row with the value
-        // specified by the data prop.
-        let cellClassName = getClassName(column, sortBy, row, columns);
 
-        let cellValue = row[column.prop];
+    let rowCells = columns.map((column, index) => {
+      let cellClassName = getClassName(column, sortBy, row, columns.length);
```
mlunoe: Why did this change to `columns.length`? It seems the argument is still called `columns` in the `getClassName` function declaration

src/Table/Table.js
```diff
@@ -167,15 +184,18 @@ export default class Table extends React.Component {
...
+      let cellValue = row[column.prop];
+      let cellID;
 
+      if (!column.dontCache) {
+        cellID = this.buildUniqueID(cellClassName, cellValue, column);
```
mlunoe: Maybe we could just pass down `column.prop`, and since we are using `column.prop` multiple places let's create a variable for it.

mlunoe:
Unrelated to this PR: Maybe we should rename `getScrollTable` to just `getTable`, since we now only have this one version.

kennyt:
i changed it to `getTBody` in another PR that's currently up

mlunoe:
Oh! great :)

src/Table/Table.js
rcorral: Why just pass `column.prop`? We need to make it as unique as possible though.

src/Table/Table.js
kennyt: alright, fixed

src/Table/Table.js
kennyt: `column.prop` will be unique enough

PR #140, opened 2016-01-06
Adding first form example, ref. DCOS-4245

mlunoe:

### DIFF
--- a/src/Form/FieldCheckbox.js
+++ b/src/Form/FieldCheckbox.js
@@ -1,8 +1,8 @@
 import classNames from 'classnames';
-
 import React from 'react';
 
 import BindMixin from '../Mixin/BindMixin';
+import ItemCheckbox from './ItemCheckbox';
 import Util from '../Util/Util';
 
 export default class FieldCheckbox extends Util.mixin(BindMixin) {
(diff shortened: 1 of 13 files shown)


mesosphere-web:
Reviewer: @kennyt

src/Form/ItemCheckbox.js
```diff
@@ -0,0 +1,116 @@
...
+
+    let indeterminate = props.indeterminate;
+    if (indeterminate != null) {
+      checkbox.indeterminate = indeterminate;
+    }
+  }
+
+  componentDidUpdate() {
+    let {props, refs} = this;
+    let checkbox = React.findDOMNode(refs.checkbox);
```
kennyt: i wonder if we should cache this dom node.

if it's being updated a lot and there are a lot of checkboxes (for example, if you had a checkbox for each row in a table) and that table updates frequently. i wonder if this would be slow or not

daiyiwhale: the checkboxes need to update to receive bulk commands from the master checkbox

src/Form/FieldCheckbox.js
```diff
@@ -10,91 +10,123 @@ export default class FieldCheckbox extends Util.mixin(BindMixin) {
     return ['handleChange'];
   }
 
-  componentDidMount() {
-    if (this.props.startValue === 'indeterminate') {
-      React.findDOMNode(this.refs.checkbox).indeterminate = true;
-    }
+  handleChange(name, modelState) {
+    let {props} = this;
```
daiyiwhale: Nice (: I'm going to start using this

src/Form/ItemCheckbox.js
kennyt: yeah i agree. just saying that we can cache the checkbox so that we don't have to search the DOM constantly when it updates (maybe under the covers React does this in a super fast way, so this could be irrelevant)

src/Form/FieldCheckbox.js
mlunoe: Yeah, @kennyt actually made me realize that this was a good use of the "properties spread" (can't remember its actual name)

kennyt: I believe the official name (supported by all the standards) is "Bracket Sandwich"

src/Form/ItemCheckbox.js
```diff
@@ -0,0 +1,116 @@
...
+
+    if (props.checked != null) {
```
kennyt: we should pull this out into a method called `updateCheckbox` or something. it's the same code twice

src/Form/FieldCheckbox.js
```diff
@@ -10,91 +10,123 @@ export default class FieldCheckbox extends Util.mixin(BindMixin) {
...
-  render() {
-    let attributes;
-    let startValue = this.props.startValue;
-
-    let labelClass = classNames({
-      'form-row-element form-element-checkbox': true,
-      [this.props.labelClass]: true
+  getItems() {
+    let {props, handleChange} = this;
+    return props.startValue.map((attributes, index) => {
```
daiyiwhale: fat arrow

src/Form/FieldCheckbox.js
daiyiwhale: yum

mlunoe: :hamburger:

src/Form/ItemCheckbox.js
mlunoe: Done

src/Util/Util.js
```diff
@@ -583,19 +556,20 @@ const Util = {
     return string.charAt(0).toUpperCase() + string.slice(1, string.length);
   },
 
-  isArray: function (arg) {
-    return Object.prototype.toString.call(arg) === '[object Array]';
-  },
-
-  // Add external lodash functions
+  // Add external custom function
```
kennyt: function`s`

kennyt:
after the one minor comment i say :ship:

src/Form/FieldInput.js
```diff
@@ -26,27 +26,32 @@ export default class FieldInput extends React.Component {
   }
 
   handleEvent(event, eventObj) {
-    this.props.handleEvent(
-      event, this.props.name, eventObj.target.value, eventObj
-    );
+    let {props} = this;
+    props.handleEvent(event, props.name, eventObj.target.value, eventObj);
```
rcorral: I like `let {props} = this;`
but when we only make use of `props` once, it just seems silly to assign a variable to a variable without modifying to only use it once.

src/Form/ItemCheckbox.js
```diff
@@ -0,0 +1,110 @@
...
+  get methodsToBind() {
+    return ['handleChange'];
+  }
+
+  componentDidMount() {
+    this.updateCheckbox();
+  }
+
+  componentDidUpdate() {
+    this.updateCheckbox();
```
rcorral: These lifecycle methods don't call `super`

PR #141, opened 2016-01-06
add dontCache to Table proptypes

daiyiwhale:
Currently it is a bit of an easter egg (':

### DIFF
--- a/src/Table/Table.js
+++ b/src/Table/Table.js
@@ -324,6 +324,8 @@ Table.propTypes = {
         PropTypes.func
       ]),
       defaultContent: PropTypes.string,
+      // Column cached by default; disable with this setting.
+      dontCache: PropTypes.bool,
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @rcorral

src/Table/Table.js
```diff
@@ -324,6 +324,8 @@ Table.propTypes = {
         PropTypes.func
       ]),
       defaultContent: PropTypes.string,
+      // Column cache by default; disable with this setting.
```
kennyt: `Column cached`

after that :ship:

daiyiwhale: fix'd!

daiyiwhale: I like how I managed to make a grammatical mistake within 2 lines of code |:

PR #142, opened 2016-01-07
Upgrade Babel@6

rcorral:
Changed the way Util.mixin works. Because we use webpack, it is not aware of the prototypal inheritance that we dynamically try to build. So therefore using `super` doesn't work as expected.


mesosphere-web:
Reviewer: @mlunoe

src/Confirm/Confirm.js
```diff
@@ -95,3 +95,5 @@ Confirm.propTypes = {
   rightButtonClassName: PropTypes.string,
   rightButtonCallback: PropTypes.func.isRequired
 };
+
+module.exports = Confirm;
```
rcorral: Have to do this now when exporting. Apparently Babel 5 was doing it wrong and they've changed the way that `export default` works, so this is why there's many files that have this change.

src/Util/DOMUtil.js
```diff
@@ -1,4 +1,4 @@
-let computeInnerBound = (compstyle, acc, key) => {
+var computeInnerBound = function (compstyle, acc, key) {
```
rcorral: Made this change because tests just weren't passing for me. ¯\_(ツ)_/¯

src/Util/__tests__/Util-test.js
```diff
@@ -68,7 +68,7 @@ describe('Util', function () {
     });
 
     it('doesn\'t exclude any properties', function () {
-      let expectedResult = {
+      var expectedResult = {
```
rcorral: Same thing here, tests weren't passing because of the `let`s.
¯\_(ツ)_/¯

.webpack.config.js
```diff
@@ -0,0 +1,27 @@
...
+module.exports = {
+  devtool: 'source-map',
+  entry: glob.sync(config.files.docs.srcJS),
+  output: {filename: config.files.docs.distJS},
+  module: {
+    loaders: [{
+      exclude: /node_modules/,
+      loader: 'babel-loader',
+      query: {
+        cacheDirectory: true,
```
kennyt: nice! the previous gulpfile didn't have any caching at all.

docs/gulpTasks.js
```diff
@@ -112,14 +115,14 @@ gulp.task('docs:webpack', ['docs:eslint'], function (callback) {
       timing: true
     }));
 
-    browserSync.reload();
```
kennyt: why do we no longer do `reload` in webpack update?

kennyt:
:+1:

docs/gulpTasks.js
```diff
@@ -29,15 +33,18 @@ gulp.task('docs:browsersync', function () {
   });
 });
 
-gulp.task('docs:eslint', ['eslint'], function () {
-  return gulp.src([config.dirs.docs.srcJS + '/**/*.?(js|jsx)'])
+// Create a function so we can use it inside of webpack's watch function.
+function eslintFn () {
```
mlunoe: #nit There is an extra space here

docs/gulpTasks.js
```diff
@@ -69,11 +80,15 @@ gulp.task('docs:minify-js', ['docs:replace-js-strings'], function () {
     .pipe(gulp.dest(config.dirs.docs.distJS));
 });
 
-gulp.task('docs:replace-js-strings', ['docs:webpack'], function () {
+function replaceJsStringsFn () {
```
mlunoe: #nit There is an extra space here

PR #143, opened 2016-01-07
Adding call to `super`

mlunoe:


src/Form/ItemCheckbox.js
```diff
@@ -10,10 +10,12 @@ export default class ItemCheckbox extends Util.mixin(BindMixin) {
   }
 
   componentDidMount() {
+    super.componentDidMount();
     this.updateCheckbox();
   }
 
   componentDidUpdate() {
+    super.componentDidUpdate();
```
rcorral: Need to pass arguments.
https://facebook.github.io/react/docs/component-specs.html#updating-componentdidupdate

PR #147, opened 2016-01-07
Fix broken tasks.

rcorral:


mlunoe:
Tests were failing after this got merged

PR #153, opened 2016-01-08
Fixing npm tasks

rcorral:


mesosphere-web:
Reviewer: @kennyt

.webpack.config.js
```diff
@@ -23,5 +21,5 @@ module.exports = {
   resolve: {
     extensions: ['', '.js']
   },
-  watch: true
+  watch: process.env.NODE_ENV === "development"
```
mlunoe: single quotes

docs/gulpTasks.js
```diff
@@ -16,8 +16,12 @@ var config = require('../.build.config');
 var packageInfo = require('../package');
 var webpackConfig = require('../.webpack.config');
 
+var development = process.env.NODE_ENV === "development";
```
mlunoe: single quotes

package.json
```diff
@@ -69,7 +70,7 @@
     "dist-src": "cp -r ./src ./lib && find ./lib -name '__tests__' -type d -exec rm -r '{}' + && ./node_modules/.bin/babel lib --out-dir lib",
     "livereload": "NODE_ENV='development' ./node_modules/.bin/gulp docs:livereload",
     "npm-publish": "npm publish ./",
-    "test": "jest",
+    "test": "mv .babelrc .babelrc-tmp && jest && mv .babelrc-tmp .babelrc",
```
mlunoe: If one test fails, this will continue to fail until the `.babelrc` is moved back

rcorral: 1. We shouldn't be merging things in that don't pass tests. So that is a small worry for me.
2. This is only temporary. See [this](https://github.com/mesosphere/reactjs-components/commit/19662452dee1302fd6e742b6fc4799decc9a413e)

docs/gulpTasks.js
rcorral: Done.

.webpack.config.js
rcorral: Done.

package.json
mlunoe: Ok, so for others. When this gets merged and you are developing on tests:
1. Delete the `.bashrc`
2. Create your tests running `./node_modules/.bin/jest`
3. Add the `.bashrc` back and continue with your development

The reason for this is that running the tests will look for the `.bashrc`, but because it runs on a different version of babel, it will fail with the current `.bashrc`.

@rcorral is working on updating jest to use the new babel version

PR #158, opened 2016-01-12
clarify table prop doc

daiyiwhale:
```
  // Optional selector of the parent element that has a scrollbar in order to
  // listen to its scroll event.
  containerSelector: PropTypes.string
```

### DIFF
--- a/docs/src/Table/index.js
+++ b/docs/src/Table/index.js
@@ -228,20 +228,36 @@ class TableExample extends React.Component {
   // Define how columns should be rendered and if they are sortable.
   columns: PropTypes.arrayOf(
     PropTypes.shape({
+      // Attributes to pass down to each column item.
       attributes: PropTypes.object,
+
+      // Class to give to each column item.
+      // Can be a function to programmatically create a class.
       className: PropTypes.oneOfType([
         PropTypes.string,
         PropTypes.func
       ]),
+
+      // Content to default to in the case of no data for the prop.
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @rcorral

kennyt:
can you give an example? for example if the container is
```html
<div class="container">
  <Table />
</div>
```
then the containerSelector would have to be `.container`

rcorral:
+1 for example so that it's clear

kennyt:
bump

rcorral:
ping @daiyiwhale

daiyiwhale:
[rises from the dead]

docs/src/Table/index.js
```diff
@@ -261,44 +261,40 @@ class TableExample extends React.Component {
...
   colGroup: PropTypes.object,
 
-  // Settings for each column in table.
+  // Define how columns should be rendered and if they are sortable.
   columns: PropTypes.arrayOf(
     PropTypes.shape({
-      // Attributes to pass down to each column item.
-      attributes: React.PropTypes.object,
-
-      // Class to give to each column item. Can be a function to programmatically create a class.
```
mlunoe: Maybe we should add these comments back into the table instead of removing them from here, so they are still identical, but still provide useful information

mlunoe:
After my comment :shipit:

daiyiwhale:
prop comments are back :+1:

PR #159, opened 2016-01-13
Allow a custom function to be passed into showLabel

kennyt:
the `showLabel` prop inside of definition can now have three possible value types:

* **boolean**: if true, will show the field name as a label. if false, will show nothing.
* **string**: will use string as label.
* **function**: will invoke and return the result of the function

### DIFF
--- a/docs/src/Form/index.js
+++ b/docs/src/Form/index.js
@@ -49,7 +49,7 @@ class FormExample extends React.Component {
             checked: false
           }
         ],
-        description: 'What is your role?',
+        showLabel: 'What is your role?',
         name: 'role',
         validation: function (value) {
           let result = false;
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @jfurrow

daiyiwhale:
Don't forget to update the description of `showLabel` https://github.com/mesosphere/reactjs-components/pull/159/files#diff-56a980b3f33a388ae085cd222f0fa748R186 and also in the doc!

rcorral:
@kennyt why does it need to be a function? Can't it just be jsx?

kennyt:
@rcorral you're right. fixed

updated the docs too!

kennyt:
alright, I redid it

daiyiwhale:
Mysteriously, tests now failing?

kennyt:
all tests currently failing with the same errors

kennyt:
alright, updated the proptypes too

daiyiwhale:
:+1:

PR #160, opened 2016-01-13
Allow modals to get bigger as the viewport gets bigger

kennyt:
Currently, if the modals are capped out at the `maxHeight` and the maxHeight gets bigger, the modal doesn't also get bigger. This fixes that.


mesosphere-web:
Reviewer: @daiyiwhale

daiyiwhale:
Nice!

rcorral:
I'm stoked for this.

PR #162, opened 2016-01-13
Don’t show a single item when table not on screen

kennyt:
Was trying to be clever before and only show 1 item when table was no on screen. This caused problems because if the container wasn't the window, it wouldn't refresh and show the full items when it was on the screen.

This fix will now show the necessary items whether the table is on the screen or not

### DIFF
--- a/docs/src/Table/index.js
+++ b/docs/src/Table/index.js
@@ -545,19 +545,17 @@ class TableExample extends React.Component {
                     </p>
                   </div>
                 </div>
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @mlunoe

daiyiwhale:
The rows are now loading, but I can't scroll within the container.

daiyiwhale:
Looks good!

PR #164, opened 2016-01-14
Add textarea field type

jfurrow:
This adds the ability to render a `textarea` with the same behavior as other `input` elements.


mesosphere-web:
Reviewer: @kennyt

src/Form/FieldTypes.js
```diff
@@ -1,11 +1,13 @@
...
 import FieldInput from './FieldInput';
 import FieldPassword from './FieldPassword';
+import FieldTextarea from './FieldTextarea';
 
 const FieldTypes = {
   'checkbox': FieldCheckbox,
+  'password': FieldPassword,
   'text': FieldInput,
-  'password': FieldPassword
+  'textarea': FieldTextarea
```
jfurrow: :abcd:

src/Form/FieldTextarea.js
```diff
@@ -0,0 +1,42 @@
...
+
+export default class FieldTextarea extends FieldInput {
+  getInputElement(attributes) {
+    let {props} = this;
+
+    let classes = classNames(props.inputClass, props.sharedClass);
+    attributes = this.bindEvents(attributes);
+
+    if (this.isEditing() || props.writeType === 'input') {
+      return (
```
kennyt: did we want to also wrap this inside of a `.render` call or no?

mlunoe: Also aren't we repeating FieldInput here?
Could we make it a bit more flexible by adding the option of rendering different tags (when editing) for a core text field component and then extending that to create a FieldInput and FieldTextarea.
Just an idea

jfurrow: @kennyt Yeah, I'll add the render wrapper here too.

@mlunoe You are right, but I am concerned about doing this in the time frame we have... Would you be okay with me creating a JIRA task for this, which can be completed when the installer is further along? Others, feel free to weigh in here.

kennyt:
alright this looks good to me

src/Form/FieldTextarea.js
```diff
@@ -0,0 +1,49 @@
+import classNames from 'classnames';
+/* eslint-disable no-unused-vars */
+import React from 'react';
+/* eslint-enable no-unused-vars */
+
+import FieldInput from './FieldInput';
+import IconEdit from './icons/IconEdit';
+
+export default class FieldTextarea extends FieldInput {
```
rcorral: `module.exports =`

rcorral:
looks good after that comment

jfurrow:
@rcorral Done!

PR #167, opened 2016-01-14
Allow react elements in definition, allow custom render functions

jfurrow:
This introduces two features:
* Allows users to pass a custom `render` function for each form element. The custom `render` function receives the input element and is expected to return a transformed element, or the same element, in its place.
* Allows users the ability to pass components as items in the `definition` array, which will not be modified in any way by the component.

### DIFF
--- a/src/Form/FieldInput.js
+++ b/src/Form/FieldInput.js
@@ -135,8 +135,8 @@ export default class FieldInput extends React.Component {
       );
     }
 
-    if (this.props.render) {
-      return this.props.render(inputContent);
+    if (this.props.renderer) {
+      return this.props.renderer(inputContent);
     }
 
     return inputContent;
@@ -191,7 +191,7 @@ FieldInput.propTypes = {
   // (usually passed down from form definition)
   name: React.PropTypes.string.isRequired,
   // Custom render function, receives the input element as its only argument
-  render: React.PropTypes.func,
+  renderer: React.PropTypes.func,
   // Optional boolean, string, or react node.
   // If boolean: true - shows name as label; false - shows nothing.
   // If string: shows string as label.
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @mlunoe

kennyt:
I think this looks good :ship:

src/Form/FieldInput.js
```diff
@@ -183,6 +190,8 @@ FieldInput.propTypes = {
   // Name of the field property
   // (usually passed down from form definition)
   name: React.PropTypes.string.isRequired,
+  // Custom render function, receives the input element as its only argument
+  render: React.PropTypes.func,
```
rcorral: What if the property name was `renderer`? Would that make more sense?

src/Form/Form.js
```diff
@@ -191,6 +191,9 @@ class Form extends Util.mixin(BindMixin) {
   buildStateObj(definition, fieldProp) {
     let stateObj = {};
     Util.flatten(definition).forEach(function (formControlOption) {
+      if (React.isValidElement(formControlOption)) {
```
rcorral: Maybe a comment could be useful here in the future

src/Form/Form.js
```diff
@@ -227,6 +230,14 @@ class Form extends Util.mixin(BindMixin) {
     ]);
 
     return definition.map((formControlOption, i) => {
+      if (React.isValidElement(formControlOption)) {
```
rcorral: Also a comment here as to why we're wrapping the `formControlOption` is it so it has a `key`?

src/Form/FieldInput.js
jfurrow: `renderer` sounds good to me, done!

src/Form/Form.js
jfurrow: Done!

src/Form/Form.js
jfurrow: Done!

PR #168, opened 2016-01-15
Remove default password

jfurrow:
This removes the default password string, which displayed the same string for password fields whenever they were not in focus.

By default it will behave as a regular password input field, where it's empty to start and then displays the obfuscated password as the user enters it. On blur, the obfuscated password remains.

Users can specify a `defaultPasswordValue` which will be displayed when a password field does not have focus, in order to hide the length of the actual password in the input field.

In this example, the `defaultPasswordValue` is set to a two-character string:

### DIFF
--- a/src/Form/FieldPassword.js
+++ b/src/Form/FieldPassword.js
@@ -6,8 +6,6 @@ import React from 'react';
 import FieldInput from './FieldInput';
 import IconEdit from './icons/IconEdit';
 
-const DEFAULT_PASSWORD_TEXT = '••••••';
-
 const METHODS_TO_BIND = ['handleOnFocus'];
 
 export default class FieldPassword extends FieldInput {
@@ -34,10 +32,7 @@ export default class FieldPassword extends FieldInput {
     attributes = this.bindEvents(attributes, this.props.handleEvent);
     attributes.onFocus = this.handleOnFocus;
 
-    let startValue = DEFAULT_PASSWORD_TEXT;
-    if (this.focused) {
-      startValue = attributes.startValue;
-    }
+    let startValue = attributes.startValue || '';
 
     if (this.isEditing() || this.props.writeType === 'input') {
       inputContent = (
--- a/src/Form/FieldPassword.js
+++ b/src/Form/FieldPassword.js
@@ -6,7 +6,7 @@ import React from 'react';
 import FieldInput from './FieldInput';
 import IconEdit from './icons/IconEdit';
 
-const METHODS_TO_BIND = ['handleOnFocus'];
+const METHODS_TO_BIND = ['handleOnBlur', 'handleOnFocus'];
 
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @rcorral

kennyt:
You say that they can still specify the `defaultPasswordValue`, but I don't see it being used anywhere

jfurrow:
You're right, looks like I broke that...

jfurrow:
@kennyt I fixed the `defaultPasswordValue` behavior

src/Form/FieldPassword.js
```diff
@@ -31,11 +38,14 @@ export default class FieldPassword extends FieldInput {
...
+    attributes.onBlur = this.handleOnBlur;
     attributes.onFocus = this.handleOnFocus;
 
-    let startValue = DEFAULT_PASSWORD_TEXT;
+    let fieldValue = attributes.defaultPasswordValue ||
+      attributes.startValue || '';
+
     if (this.focused) {
-      startValue = attributes.startValue;
+      fieldValue = attributes.startValue;
```
rcorral: If we're focused on the field, then we want to show the `startValue`? I'm not sure that makes sense.
If I want to start typing on the field, and the `startValue` is `foo` everytime I focus it'll use that value?

src/Form/FieldPassword.js
```diff
@@ -31,11 +38,14 @@ export default class FieldPassword extends FieldInput {
   getInputElement(attributes) {
     let classes = classNames(this.props.inputClass, this.props.sharedClass);
     attributes = this.bindEvents(attributes, this.props.handleEvent);
+    attributes.onBlur = this.handleOnBlur;
     attributes.onFocus = this.handleOnFocus;
 
-    let startValue = DEFAULT_PASSWORD_TEXT;
+    let fieldValue = attributes.defaultPasswordValue ||
+      attributes.startValue || '';
```
rcorral: @kennyt why don't we have `defaultProps` or `proptypes`?

src/Form/FieldPassword.js
jfurrow: `attributes.startValue` actually gets modified by the `Form` component, so it will always be what the user most recently edited.

rcorral: why is it prefixed with `start` then?

jfurrow: I'm not sure, I didn't change this from the original implementation. But I can if you'd like.

src/Form/FieldPassword.js
kennyt: I'll add those in another PR

kennyt:
:ship:

rcorral:
konflicts

jfurrow:
@rcorral fixed!

PR #169, opened 2016-01-15
Turn table cell caching off by default

kennyt:
Turning caching on by default makes the table harder to use up front + caching is only for niche cases.

### DIFF
--- a/src/Table/Table.js
+++ b/src/Table/Table.js
@@ -179,7 +179,7 @@ class Table extends React.Component {
       let cellValue = row[prop];
       let cellID;
 
-      if (!column.dontCache) {
+      if (column.cacheCell === true) {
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

daiyiwhale:
I think this makes a lot more sense :D

daiyiwhale:
You might have to rerun those tests though

src/Table/Table.js
```diff
@@ -200,7 +200,7 @@ class Table extends React.Component {
           </td>
         );
 
-        if (column.dontCache) {
+        if (!column.cacheCell) {
```
rcorral: If someone doesn't define this prop, will it be a problem as this will conditional will be truthy?

PR #173, opened 2016-01-19
Extend handleEvent to pass more event information

jfurrow:
For each event, `handleEvent` will pass the `fieldName` and `fieldValue` of the element that triggered the event, as well as the event object itself.

### DIFF
--- a/src/Form/FieldPassword.js
+++ b/src/Form/FieldPassword.js
@@ -23,7 +23,7 @@ export default class FieldPassword extends FieldInput {
       this.forceUpdate();
     }
 
-    this.props.handleEvent('blur', this.props.name, event.target.value);
+    this.props.handleEvent('blur', this.props.name, event.target.value, event);
   }
 
   handleOnFocus(event) {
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @rcorral

daiyiwhale:
nice :+1:

kennyt:
Can you change it in `FieldInput` also?

jfurrow:
I think it's already set in `FieldInput` unless I'm missing something: `props.handleEvent(event, props.name, eventObj.target.value, eventObj);`

src/Form/FieldPassword.js
```diff
@@ -32,7 +32,7 @@ export default class FieldPassword extends FieldInput {
       this.forceUpdate();
     }
 
-    this.props.handleEvent('focus', this.props.name, event);
+    this.props.handleEvent('focus', this.props.name, event.target.value);
```
kennyt: you're not passing the event are you?

kennyt:
you're right. `FieldInput` is good

src/Form/FieldPassword.js
jfurrow: Woops, good catch. :sleeping:

PR #174, opened 2016-01-20
Expose scrollbar and height adjustment in Modal

daiyiwhale:
Changes to Modal: allow disabling of gemini scrollbar and allow disabling of dynamic height adjustment. Disabling scrollbar allows for components to overflow the modal (such as dropdown).

```
<Modal
  dynamicHeight={false}
  useGemini={false}
  ... >
```

### DIFF
--- a/src/Modal/ModalContents.js
+++ b/src/Modal/ModalContents.js
@@ -282,17 +282,11 @@ class ModalContents extends Util.mixin(BindMixin) {
     }
 
     let calculatedHeight = this.heightInfo;
-
     let contentHeight = calculatedHeight.contentHeight;
-    let modalStyle = {height: calculatedHeight.height};
+    let modalStyle = null;
 
-    let useScrollbar = this.props.useScrollbar;
-    if (useScrollbar === null) {
-      if (calculatedHeight.height !== 'auto') {
-        useScrollbar = true;
-      } else {
-        useScrollbar = false;
-      }
+    if (this.props.dynamicHeight) {
+      modalStyle = {height: calculatedHeight.height};
     }
 
     return (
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/Modal/ModalContents.js
```diff
@@ -286,9 +286,13 @@ class ModalContents extends Util.mixin(BindMixin) {
     let contentHeight = calculatedHeight.contentHeight;
     let modalStyle = {height: calculatedHeight.height};
 
-    let useScrollbar = false;
-    if (calculatedHeight.height !== 'auto') {
-      useScrollbar = true;
+    let useScrollbar = this.props.useScrollbar;
+    if (useScrollbar === null) {
```
jfurrow: We should use double equals here to catch an `undefined` value as well.

jfurrow:
Looks good to me after my comment!

src/Modal/ModalContents.js
```diff
@@ -282,13 +282,11 @@ class ModalContents extends Util.mixin(BindMixin) {
...
 
     let calculatedHeight = this.heightInfo;
-
     let contentHeight = calculatedHeight.contentHeight;
-    let modalStyle = {height: calculatedHeight.height};
+    let modalStyle = null;
 
-    let useScrollbar = false;
-    if (calculatedHeight.height !== 'auto') {
-      useScrollbar = true;
```
rcorral: I think you still want to use this `if` statement, you just want to allow override from the prop. Like this:
```
let useScrollbar = false;
if (this.props.useGemini && calculatedHeight.height !== 'auto') {
  useScrollbar = true;
}
```

src/Modal/ModalContents.js
```diff
@@ -282,13 +282,11 @@ class ModalContents extends Util.mixin(BindMixin) {
...
+    if (this.props.dynamicHeight) {
+      modalStyle = {height: calculatedHeight.height};
```
rcorral: Could we add a test for this?

src/Modal/ModalContents.js
rcorral: Also add a test please!

daiyiwhale:
Tests in!

rcorral:
Looks good! I say :ship:

kennyt:
lgtm

PR #175, opened 2016-01-20
Add eventType to eventObject that gets passed to onChange handler

jfurrow:
The `eventType` (blur, focus, change) needs to be passed back to the `onChange` handler.


mesosphere-web:
Reviewer: @kennyt

PR #183, opened 2016-02-01
DCOS-2044: Add tooltip component

jfurrow:
The tooltip component has the concept of `position` and `anchor`. `position` refers to the position of the tooltip relative to its trigger (which is its children), and acceptable values are `top`, `right`, `bottom`, and `left`. `anchor` refers to where the tooltip is anchored on the triggered element, and acceptable values are `start`, `center`, and `end`. Think of flexbox alignment when thinking about the `start` and `end` values.

### DIFF
--- a/docs/src/Tooltip/index.js
+++ b/docs/src/Tooltip/index.js
@@ -60,14 +60,14 @@ class ToolTipExample extends React.Component {
                   <Tooltip content="I'm a tooltip!">
                     <button className="button">Top</button>
                   </Tooltip>
-                  <Tooltip content="I'm a tooltip!" position="right">
-                    <button className="button">Right</button>
-                  </Tooltip>
                   <Tooltip content="I'm a tooltip!" position="bottom">
                     <button className="button">Bottom</button>
                   </Tooltip>
                   <Tooltip content="I'm a tooltip!" position="left">
                     <button className="button">Left</button>
+                  </Tooltip>
+                  <Tooltip content="I'm a tooltip!" position="right">
+                    <button className="button">Right</button>
                   </Tooltip>
                 </div>
               </section>
--- a/docs/src/Tooltip/index.js
+++ b/docs/src/Tooltip/index.js
@@ -1,0 +1,120 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import Tooltip from '../../../src/Tooltip/Tooltip.js';
+
+class ToolTipExample extends React.Component {
+  render() {
+    return (
+      <div className="row canvas-pod">
+        <div>
+          <h2 className="short-bottom">Tooltips</h2>
+          <p>
+            A tooltip appears when a user hover overs an element. The tooltips will adjust their position if the supplied positioning results in the tooltip rendering outside of the veiwport. For example, if you supply the position of "bottom", but the element is at the bottom of the viewport, it will render the tooltip at the top of the element.
+          </p>
+          <p>
+            The tooltip allows users to set the position and anchor. Position refers to the position of the tooltip relative to its trigger (which is its children), and acceptable values are top, right, bottom, and left. Anchor refers to where the tooltip is anchored on the triggered element, and acceptable values are start, center, and end. Think of flexbox alignment when thinking about the start and end values.
+          </p>
+          <p>
+            View component source <a href="https://github.com/mesosphere/reactjs-components/blob/master/src/Tooltip/Tooltip.js">here</a>.
+            View full example source <a href="https://github.com/mesosphere/reactjs-components/blob/master/docs/src/Tooltip/index.js">here</a>.
+          </p>
+          <h3>Properties API</h3>
+          <div className="example-block">
+            <pre className="prettyprint linenums flush-bottom">
+{`Tooltip.propTypes = {
+  // Anchor the tooltip to an edge of the tooltip trigger. Start and end refer
+  // to the logical respective edges of the tooltip. When tooltip is positioned
+  // on the right, start refers to the top of the tooltip. When tooltip is
+  // positioned on the bottom, start refers to the left edge of the tooltip.
+  anchor: React.PropTypes.oneOf(['start', 'center', 'end']),
+  // The children will be used as the trigger.
+  children: React.PropTypes.node.isRequired,
+  className: React.PropTypes.string,
+  // The tooltip's content.
+  content: React.PropTypes.node.isRequired,
+  // Position the tooltip on an edge of the tooltip trigger. Default is top.
+  position: React.PropTypes.oneOf(['top', 'bottom', 'right', 'left']),
+  tooltipWrapperClassName: React.PropTypes.string,
+  // Explicitly set the width of the tooltip. Default is auto.
+  width: React.PropTypes.number,
+  // Allow the text content to wrap. Default is false. This should be used with
+  // the width property, because otherwise the width of the content will be the
+  // same as the trigger.
+  wrapText: React.PropTypes.bool
+};`}
+            </pre>
+          </div>
+          <div className="example-block flush-bottom example-block-overflow">
+            <div className="example-block-content">
+              <section className="row canvas-pod">
+                <div className="column-12">
+                  <p>
+                    Hover over the following buttons to observe the behavior
+                    of the tooltips.
+                  </p>
+                </div>
+              </section>
+              <section className="row canvas-pod">
+                <div className="column-12 column-mini-12">
+                  <Tooltip content="I'm a tooltip!">
+                    <button className="button">Top</button>
+                  </Tooltip>
+                  <Tooltip content="I'm a tooltip!" position="right">
+                    <button className="button">Right</button>
+                  </Tooltip>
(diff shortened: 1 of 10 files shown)


mesosphere-web:
Reviewer: @rcorral

src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,141 @@
...
+  // Anchor the tooltip to an edge of the tooltip trigger. Start and end refer
+  // to the logical respective edges of the tooltip. When tooltip is positioned
+  // on the right, start refers to the top of the tooltip. When tooltip is
+  // positioned on the bottom, start refers to the left edge of the tooltip.
+  anchor: React.PropTypes.oneOf(['start', 'center', 'end']),
+  // The children will be used as the trigger.
+  children: React.PropTypes.node.isRequired,
+  className: React.PropTypes.string,
+  // The tooltip's content.
+  content: React.PropTypes.node.isRequired,
```
kennyt: can content be a string? Isn't it usually a string and not a node?

jfurrow: It can be both, because sometimes we want to be able to pass in HTML (like in the Installer UI we have some links)

kennyt: so will this throw a warning if we pass it a string?

jfurrow: No, PropTypes.node accepts all kinds of junk:
```javascript
    // Anything that can be rendered: numbers, strings, elements or an array
    // (or fragment) containing these types.
    optionalNode: React.PropTypes.node,
```

src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,141 @@
...
+  'handleMouseLeave'
+];
+
+class Tooltip extends React.Component {
+  constructor() {
+    super();
+
+    this.state = {isOpen: false};
+
+    METHODS_TO_BIND.forEach((method) => {
```
kennyt: There's a super useful `BindMixin` in this repo

src/Tooltip/Tooltip.js
kennyt: wtf

daiyiwhale: Amazing

src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  getIdealPosition() {
+    // Get the anchor and position from state if possible. If not, get it from
+    // the props.
+    let anchor = this.state.anchor || this.props.anchor;
+    let position = this.state.position || this.props.position;
+
+    if (this.refs.tooltipContent) {
+      let viewportHeight = global.window.innerHeight;
```
kennyt: There's a `DOMUtil#getViewportHeight` that you can use. You can also create a `#getViewportWidth` too

docs/src/Tooltip/index.js
```diff
@@ -0,0 +1,120 @@
...
+                  </p>
+                </div>
+              </section>
+              <section className="row canvas-pod">
+                <div className="column-12 column-mini-12">
+                  <Tooltip content="I'm a tooltip!">
+                    <button className="button">Top</button>
+                  </Tooltip>
+                  <Tooltip content="I'm a tooltip!" position="right">
+                    <button className="button">Right</button>
```
daiyiwhale: This is kind of silly but maybe we could switch the right and left tooltips, because when hovering over buttons left-to-right or right-to-left the tooltips get in the way of the next button and you have to mouseoff one button to get to the next button.

src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,141 @@
...
+      let viewportWidth = global.window.innerWidth;
+      let tooltipContent = ReactDOM.findDOMNode(this.refs.tooltipContent);
+      let tooltipPosition = tooltipContent.getBoundingClientRect();
+
+      // Change the position if the tooltip will be rendered off the screen.
+      if (position === 'right' && tooltipPosition.right > viewportWidth) {
```
kennyt: maybe you can break this method up more by pulling the getting position logic into a `#getPostion` and the getting anchor logic into a `#getAnchor`

docs/src/Tooltip/index.js
kennyt: great catch. I agree

docs/src/index.less
```diff
@@ -13,7 +13,8 @@
 @import "../../src/List/list.less";
 @import "../../src/Modal/modal.less";
 @import "../../src/SidePanel/side-panel.less";
-@import "../../src/Table/table.less";
+@import "../../src/SidePanel/side-panel.less";
+@import "../../src/Tooltip/tooltip.less";
```
daiyiwhale: Does this also need to be imported into `src/index.less`?

src/Tooltip/Tooltip.js
mlunoe: :star:

docs/src/Tooltip/index.js
mlunoe: :+1:

src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,141 @@
...
+        tooltipPosition.bottom > viewportHeight) {
+        position = 'top';
+      }
+
+      // Change the anchor if the tooltip will be rendered off the screen.
+      if (position === 'right' || position === 'left') {
+        if (tooltipPosition.top < 0) {
+          anchor = 'start';
+        } else if (tooltipPosition.bottom > viewportHeight) {
+          anchor = 'end';
```
mlunoe: Could we default the anchor to be 'end' or 'start'? I believe then you can remove the nested if-statements and just have a if-else.

src/Tooltip/Tooltip.js
mlunoe: :+1:

src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,141 @@
...
+  handleMouseLeave() {
+    this.setState({isOpen: false});
+  }
+
+  render() {
+    let {props, state} = this;
+
+    // Get the anchor and position from state if possible. If not, get it from
+    // the props.
+    let anchor = state.anchor || props.anchor;
```
mlunoe: Can't we derive the `anchor` and `position` every time, so we don't have duplicate information in state/props?

src/Tooltip/Tooltip.js
mlunoe: Yup, I've noticed that we have `node` or `string` or whatever options elsewhere. If it is something to render, then `node` should be fine :)

src/Tooltip/Tooltip.js
jfurrow: We can't use any function in `#render` that calls `#findDOMNode`, see the error from React:

src/Tooltip/Tooltip.js
jfurrow: Done!

src/Tooltip/Tooltip.js
jfurrow: Done!

docs/src/Tooltip/index.js
jfurrow: Good call! I rearranged the buttons to be in this order: top, bottom, left, right

I think that solves the problem and maintains some logical ordering.

src/Tooltip/Tooltip.js
jfurrow: Done!

docs/src/index.less
jfurrow: Probably! Done!

src/Tooltip/Tooltip.js
jfurrow: Yay React!

src/Tooltip/Tooltip.js
jfurrow: I think the vast majority of users will want the tooltip anchor to be `center`, so I feel like we should keep that as the default. I'll try to break this `if` up more reasonably.

jfurrow: This reduces the number of lines a little bit, but it's more difficult to read in my opinion:
```javascript
let horizontalPosition = position === 'right' || position === 'left';
    let verticalPosition = position === 'top' || position === 'bottom';

    if ((horizontalPosition && renderedPosition.top < 0)
      || (verticalPosition && renderedPosition.right > viewportWidth)) {
      anchor = 'start';
    } else if ((horizontalPosition && renderedPosition.bottom > viewportHeight)
      || (verticalPosition && renderedPosition.left < 0)) {
      anchor = 'end';
    }
```

src/Tooltip/Tooltip.js
daiyiwhale: Yay diligence-enforced consistency!

daiyiwhale:
This is supercool! So excited :D :sparkles:

src/Tooltip/Tooltip.js
mlunoe: :+1:

PR #184, opened 2016-02-02
At super small widths, have the width property apply to all modals

kennyt:
This was causing the `Confirm` modal (which doesn't have `.modal-large` to have weird widths at small screen sizes

### DIFF
--- a/src/Modal/modal.less
+++ b/src/Modal/modal.less
@@ -121,7 +121,7 @@
 
 & when (@screen-mini-enabled) {
   @media (min-width: @screen-mini) {
-    .modal.modal-large {
+    .modal {
       width: 500px;
     }
   }


mesosphere-web:
Reviewer: @daiyiwhale

daiyiwhale:
Oh nice, mystery solved :D

mlunoe:
What visual difference does this have? I can't seem to produce a difference.
Also, the `.modal-large` classes was introduced to control the widths of the `.modal-large` components, since `.modal` already has an equal size setting from canvas.
Finally, if we need to change this, why is the `.modal-large` class only removed from the mini mediaquery?
I think I need a little bit of context here

kennyt:
go on master than open a confirm modal. then make your screen super small and you'll see the odd behavior

PR #185, opened 2016-02-02
Add option to not use gemini-scrollbar

daiyiwhale:
This PR allows you to opt out of gemini-scrollbar in Dropdown and SidePanel. Gemini-scrollbar will still be used by default.

Incidentally, scrolling on SidePanel will now not cause the background to scroll. (Edit: previously, when the sidepanel was active, scrolling would scroll to the bottom of the sidepanel contents and then begin scrolling the background page contents. I fixed it so that it's not possible to scroll the background page content while the sidepanel is active.)


mesosphere-web:
Reviewer: @jfurrow

src/SidePanel/SidePanelContents.js
```diff
@@ -16,6 +16,14 @@ class SidePanelContents extends Util.mixin(BindMixin) {
     ];
   }
 
+  componentWillReceiveProps(newProps) {
```
kennyt: We generally call this `nextProps` :)

daiyiwhale: argh that's what I get for copypasta (':

daiyiwhale: done!

docs/src/Dropdown/index.js
```diff
@@ -337,6 +337,11 @@ class DropdownExample extends React.Component {
   transition: React.PropTypes.bool,
   // The prefix of the transition classnames.
   transitionName: React.PropTypes.string,
+  // Transition lengths
+  transitionEnterTimeout: React.PropTypes.number,
+  transitionLeaveTimeout: React.PropTypes.number,
+  // Optional disable Gemini scrollbar. Defaults to true.
```
mlunoe: Maybe instead: "Option to use Gemini scrollbar. Defaults to true"

mlunoe:
What does this mean?
"Incidentally, scrolling on SidePanel will now not cause the background to scroll."

mlunoe:
Modals text is overflowing instead of scrolling for me:

daiyiwhale:
@mlunoe previously, when the sidepanel was active, scrolling would scroll to the bottom of the sidepanel contents and then begin scrolling the background page contents. I fixed it so that it's not possible to scroll the background page content while the sidepanel is active.

daiyiwhale:
Oh no sad modal ):

kennyt:
resizing the modal when `useGemini` is false also behaves weirdly

mlunoe:
I can't scroll in the sidepanel at all anymore :(

mlunoe:
Ok, side panel scroll problem was not introduced here. I created a separate PR for the fix here: https://github.com/mesosphere/reactjs-components/pull/187
(team effort with @jfurrow @daiyiwhale!)

daiyiwhale:
I'll create a separate PR for the resizing of Modal, as the issues were introduced elsewhere (https://github.com/mesosphere/reactjs-components/pull/174)

kennyt:
text overflow for modal is still a problem atm, correct?

daiyiwhale:
I have a separate pr for the modal fixes

daiyiwhale:
Ah ok, I remember now. The reason we allow it to be possible for the modal to overflow is because we wanted it to be possible for components such as Dropdown to expand past the modal borders.

daiyiwhale:
Alright, when gemini is turned off we're catching the overflow in a native scrollbar. Users will have to manually turn off overflow for now if they want stuff like Dropdown to overflow.

docs/src/Dropdown/index.js
daiyiwhale: done!

docs/styles/page-content.less
```diff
@@ -32,6 +32,9 @@ Variables
   overflow: hidden;
 }
 
+.visible-overflow {
+  overflow: auto;
+}
```
jfurrow: We don't need this class anymore, right?

daiyiwhale: oh right! I thought we did it together, but it was lost to chaos :fallen_leaf:

PR #186, opened 2016-02-02
Remove transition examples in docs

kennyt:
There are no more transitions in Tables

### DIFF
--- a/docs/src/Table/index.js
+++ b/docs/src/Table/index.js
@@ -102,12 +102,6 @@ let rows = [
 class TableExample extends React.Component {
   constructor(props) {
     super(props);
-    this.state = {
-      rowAdded: false
-    };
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @mlunoe

docs/src/Table/index.js
```diff
@@ -117,16 +111,6 @@ class TableExample extends React.Component {
     this.forceUpdate();
```
daiyiwhale: Do we still need this?

kennyt: yeah this was unrelated to the transition i believe. this had something to do with the infinite scroll table and gemini scrollbar

daiyiwhale: Weird, okay :|

PR #187, opened 2016-02-03
Fix side panel scroll

mlunoe:
This fixes a problem with the side panel, where the contents wouldn't scroll when Mac OSX scrollbar setting was set to: "Automatically based on mouse or trackpad" and a mouse or external trackpad is used.

### DIFF
--- a/src/overrides/gemini-scrollbar.less
+++ b/src/overrides/gemini-scrollbar.less
@@ -64,7 +64,7 @@
 }
 
 .container-scrollable {
-  flex: 8;
+  flex: 1;
   overflow: auto;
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

src/overrides/gemini-scrollbar.less
```diff
@@ -64,7 +64,7 @@
 }
 
 .container-scrollable {
-  flex: 8;
+  flex: 1;
```
mlunoe: There is no reason for it to be 8, it just need `flex: 1`

PR #197, opened 2016-02-23
Add ability to change row class or render differently

kennyt:


src/Form/Form.js
```diff
@@ -244,6 +244,10 @@ class Form extends Util.mixin(BindMixin) {
         );
       }
 
+      if (formControlOption.render) {
+        return formControlOption.render();
```
mlunoe: Could we add a test for this?

kennyt: done

jfurrow:
:ship:

PR #205, opened 2016-02-29
Fix dropdown bug in IE

jfurrow:


mesosphere-web:
Reviewer: @daiyiwhale

PR #210, opened 2016-03-01
Remove cursor type override

daiyiwhale:
Removing cursor type style to allow elements within table to set their own cursor types.


mesosphere-web:
Reviewer: @jfurrow

PR #217, opened 2016-03-16
Fixing IE scrollTop bug

mlunoe:
Table would not show up in some cases with the Table component


mesosphere-web:
Reviewer: @daiyiwhale

PR #231, opened 2016-03-23
Introduce submit button as a fieldtype for form

kennyt:


mesosphere-web:
Reviewer: @jfurrow

docs/src/Form/index.js
```diff
@@ -42,8 +42,9 @@ class FormExample extends React.Component {
         showLabel: 'Accept terms of service',
         required: true,
         label: 'Single checkbox',
+        value: true,
         validation: function (value) {
-          return value && value.checked;
+          return value;
```
mlunoe: Why did this change?

src/Form/FieldCheckboxMultiple.js
```diff
@@ -18,8 +18,7 @@ class FieldCheckboxMultiple extends Util.mixin(BindMixin) {
       props.name,
       {
         name: name,
-        checked: modelState.checked,
-        indeterminate: modelState.indeterminate
+        checked: modelState
```
mlunoe: Why did this change?

docs/src/Form/index.js
kennyt: because value is now whether it's checked or not

src/Form/FieldCheckboxMultiple.js
kennyt: because modelState is now the checked value

jfurrow: So `checked` can be one of `true`, `false`, or `'indeterminate'`?

kennyt: nope, indeterminate is still a property on the definition. it's just that indeterminate will come from the parent, the checkbox will never hold the value of indeterminate in state

jfurrow: ahh cool!

jfurrow:
:shipit:

PR #232, opened 2016-03-29
Improve tooltip.

jfurrow:
* Use of the `anchor` prop will now align the tooltip's arrow
at 50% of the width or height (depending on the `position`
prop), and the Tooltip's content will then be anchored
according to the prop definition.
* Allows custom DOM node to be used
* Pass all props that aren't specific to Tooltip to the node
* Remove specific margin styling

### DIFF
--- a/src/Tooltip/Tooltip.js
+++ b/src/Tooltip/Tooltip.js
@@ -94,6 +94,9 @@ class Tooltip extends Util.mixin(BindMixin) {
     // the props.
     let anchor = state.anchor || props.anchor;
     let position = state.position || props.position;
+    let DOMNode = props.nodeType;
+    // Pass along any props that aren't specific to the Tooltip.
+    let nodeProps = Util.exclude(props, Object.keys(Tooltip.propTypes));
 
     let tooltipClasses = classnames(props.className, `anchor-${anchor}`,
       `position-${position}`, {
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

src/Tooltip/Tooltip.js
```diff
@@ -94,6 +94,9 @@ class Tooltip extends Util.mixin(BindMixin) {
     // the props.
     let anchor = state.anchor || props.anchor;
     let position = state.position || props.position;
+    let DOMNode = props.nodeType;
```
kennyt: can we call this `elementTag`?

jfurrow: It has to begin with a capital letter, so are you cool with `ElementTag`?

kennyt: hmm in our `Form` we don't have it as a capital. 

https://github.com/mesosphere/reactjs-components/blob/master/src/Form/Form.js#L325

what if you don't create a variable and just do `props.elementTag`

jfurrow: Ahh nice, I like that! Done!

src/Tooltip/Tooltip.js
```diff
@@ -111,15 +113,16 @@ class Tooltip extends Util.mixin(BindMixin) {
     }
 
     return (
-      <div className={props.tooltipWrapperClassName}
+      <props.elementTag className={props.wrapperClassName}
         onMouseEnter={this.handleMouseEnter}
-        onMouseLeave={this.handleMouseLeave}>
+        onMouseLeave={this.handleMouseLeave}
+        {...nodeProps}>
```
kennyt: i like this :)

this always comes in handy

jfurrow: Yeah!

PR #233, opened 2016-03-30
Allow form-row-element to be overridden

kennyt:
All the classes are the same. There is just the ability to override classes you don't want now.

px


mesosphere-web:
Reviewer: @rcorral

jfurrow:
Nice!

PR #234, opened 2016-03-30
Allow columns to have custom value getters

kennyt:
This is useful because if the data doesn't have the prop at the top level, using `row[prop]` won't work (for example, you'd have to go inside a `resources` object to retrieve the value like so: `row.resources[prop]`.


mesosphere-web:
Reviewer: @jfurrow

jfurrow:
:ship:

PR #237, opened 2016-03-31
Use Portal for Tooltip

jfurrow:
This uses the portal and will work everywhere (even inside a container with `overflow: scroll`). That introduced a fair amount of additional logic to determine the position of the tooltip trigger and the ideal location for the tooltip content.

### DIFF
--- a/src/Tooltip/Tooltip.js
+++ b/src/Tooltip/Tooltip.js
@@ -26,18 +26,120 @@ class Tooltip extends Util.mixin(BindMixin) {
     this.state = {isOpen: false};
   }
 
-  componentDidMount() {
-    window.addEventListener('scroll', this.dismissTooltip);
-    window.addEventListener('resize', this.dismissTooltip);
+  componentWillUnmount() {
+    this.removeScrollListener();
   }
 
-  componentWillUnmount() {
-    window.removeEventListener('scroll', this.dismissTooltip);
-    window.removeEventListener('resize', this.dismissTooltip);
+  handleMouseEnter(currentAnchor, currentPosition) {
+    let {anchor, position, coordinates} = this.getIdealLocation(currentAnchor,
+      currentPosition);
+    this.setState({anchor, isOpen: true, position, coordinates});
+    this.addScrollListener();
+  }
+
+  handleMouseLeave() {
+    this.dismissTooltip();
+  }
+
+  handleTooltipMouseEnter() {
+    if (this.props.interactive) {
+      this.setState({isOpen: true});
+    }
+  }
+
+  handleTooltipMouseLeave() {
+    this.dismissTooltip();
+  }
+
+  addScrollListener() {
+    this.props.scrollContainer.addEventListener('scroll', this.dismissTooltip);
   }
 
   dismissTooltip() {
-    this.setState({isOpen: false});
+    if (this.state.isOpen) {
+      this.setState({isOpen: false});
+      this.removeScrollListener();
+    }
+  }
+
+  getAnchor(isVertical, anchor, clearance, tooltipWidth, tooltipHeight) {
+    // Calculate the ideal anchor.
+    if (isVertical) {
+      return this.transformAnchor(anchor, clearance.left, clearance.right,
+        tooltipWidth, clearance.boundingRect.width);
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @rcorral

src/Tooltip/Tooltip.js
```diff
@@ -4,91 +4,183 @@ import ReactDOM from 'react-dom';
...
+    }
+
+    return {
+      left: triggerRect.left - tooltipWidth + ARROW_SIZE,
+      top: triggerRect.top + triggerRect.height / 2
+    };
+  }
+
+  isVertical(position) {
+    if (position === 'left' || position === 'right') {
```
kennyt: we can just do `return position !== 'left' && position !== 'right'`

src/Tooltip/Tooltip.js
```diff
@@ -4,91 +4,183 @@ import ReactDOM from 'react-dom';
...
 
     return position;
   }
 
-  getIdealLocation() {
-    let anchor = this.state.anchor || this.props.anchor;
-    let position = this.state.position || this.props.position;
+  getIdealLocation(anchor, position) {
+    let isVertical = this.isVertical(position);
+    let triggerNode = ReactDOM.findDOMNode(this.refs.triggerNode);
```
kennyt: i believe `this.refs.triggerNode` will already give you the dom node in this version of react

jfurrow:
@kennyt I implemented the things we discussed on Friday, but made some modifications:
* The scroll listener will only be added when the tooltip is open, and is removed when the tooltip closes, which should prevent any performance issues.
* The user can pass in a `scrollContainer` prop, but it's expected to be a DOM node rather than a selector (like the table accepts). I chose a DOM node because many tooltips will have the same container, so I figured we could save some time by getting the DOM node in the parent component and passing it to each tooltip.
* I created a new DOMUtil called `getNodeClearance` to get the number of pixels around an element.

Let me know what you think!

src/Tooltip/Tooltip.js
```diff
@@ -4,91 +4,173 @@ import ReactDOM from 'react-dom';
...
+    this.addScrollListener();
+  }
+
+  handleMouseLeave() {
+    this.dismissTooltip();
+  }
+
+  handleTooltipMouseEnter() {
+    if (this.props.interactive) {
+      this.setState({isOpen: true});
```
kennyt: do we need to readd the listener here? or does `handleMouseLeave` never get called at this point?

jfurrow: Yeah it should be here also. Done!

kennyt:
you currently have 4 tooltip examples of non interactive. can we do 2 interactive and 2 non interactive?

jfurrow:
@kennyt Added interactive examples and added the ability to pass a string as a selector for the scroll container. I believe this covers everything!

src/Tooltip/Tooltip.js
```diff
@@ -4,91 +4,183 @@ import ReactDOM from 'react-dom';
...
+    return this.transformAnchor(anchor, clearance.top, clearance.bottom,
+      tooltipHeight, clearance.boundingRect.height);
   }
 
-  getPosition(position, renderedPosition, viewportWidth, viewportHeight) {
+  getCoordinates(position, clearance, tooltipWidth, tooltipHeight) {
+    // Calculate the coordinates of the tooltip content.
+    if (position === 'top') {
+      return {
+        left: clearance.boundingRect.left + clearance.boundingRect.width / 2,
```
kennyt: `left` is already on `clearance`, right? so you can just do `clearance.left`

kennyt:
after my one last comment this looks good :ship:

src/Tooltip/Tooltip.js
jfurrow: These are actually different values... kind of. The `top`, `right`, `bottom`, and `left` properties directly on the `clearance` object reflect the number of pixels between the edge of the viewport and the edge of the element. The `boundingRect` object contains the position of the element on the screen. So while sometimes these values are the same (`top` and `left` will be the same for both values), their intentions are different.

PR #241, opened 2016-04-01
Fix Util.sortBy to not mutate collection

### LINKED ISSUE #239: Util.sortBy shouldn't mutate the original collection
`Util.sortBy` shouldn't mutate the original collection, as this could lead to "interesting" effects. Consider the following collection and compare functions:

```
var collection = [{name:'c', value:1},{name:'t', value:3},{name:'a', value:3}];

function compareValues (a, b) {
  var delta = a.value - b.value;
  return delta / Math.abs(delta || 0);
}

function compareNames (a , b){
  return a.name.localeCompare(b.name)
}
```

Sorting it first by name (`a,c,t`) and then by value (`c,a,t`) would result in a different ordering than sorting it directly by value (`c,t,a`), this is particularly a problem as it is used within the table and might result in a different ordering after an data update.

orlandohohmeier:
Create copy of collection before sorting it to not mutate the original collection. Added test to verify correct behaviour.

/closes #239

### DIFF
--- a/src/Util/Util.js
+++ b/src/Util/Util.js
@@ -435,9 +435,9 @@ var isArray = function (arg) {
 
 function sortBy(collection, sortProp) {
   if (isFunction(sortProp)) {
-    return collection.sort(sortProp);
+    return collection.slice().sort(sortProp);
   } else {
-    return collection.sort((a, b) => {
+    return collection.slice().sort((a, b) => {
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @jfurrow

jfurrow:
Awesome!

PR #253, opened 2016-05-11
Add several options to the Tooltip component.

jfurrow:
* `suppress` to prevent the Tooltip from displaying
* `stayOpen` to keep the tooltip open after it's triggered
* `triggerTooltip` to manually trigger it from outside the component

### DIFF
--- a/src/Tooltip/Tooltip.js
+++ b/src/Tooltip/Tooltip.js
@@ -25,7 +25,7 @@ class Tooltip extends Util.mixin(BindMixin) {
   constructor() {
     super(...arguments);
     this.container = null;
-    this.state = {isOpen: false};
+    this.state = {isOpen: false, wasTriggeredClose: false};
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/Tooltip/Tooltip.js
```diff
@@ -154,6 +163,10 @@ class Tooltip extends Util.mixin(BindMixin) {
     }
   }
 
+  triggerTooltip() {
```
kennyt: how will people get access to this method in order to invoke it?

jfurrow: refs :)

src/Tooltip/Tooltip.js
```diff
@@ -243,6 +255,8 @@ Tooltip.defaultProps = {
   interactive: false,
   position: 'top',
   scrollContainer: window,
+  stayOpen: false,
+  suppress: false,
```
kennyt: can you add these two to the `propTypes` and then update the docs?

src/Tooltip/Tooltip.js
kennyt: hmm, that's good. we should definitely refactor our `Form`s in future to do that rather than the whole `getTriggerSubmit` dance

src/Tooltip/Tooltip.js
jfurrow: Done!

src/Tooltip/Tooltip.js
jfurrow: I'm all for it!

jfurrow:
@kennyt Sorry bro but I committed a few more changes since you signed off on this.

PR #255, opened 2016-05-16
Remove onKeyDownHandler on Textareas

Poltergeist:
Textareas are multiline input fields which should not be blured
if the `Enter` key is pressed.

As an example of the problem please take a look at the message text area at http://mesosphere.github.io/reactjs-components/#form


mesosphere-web:
Reviewer: @mlunoe

jfurrow:
Good catch!

PR #257, opened 2016-05-17
Add support for focused option for input fields

pierlo-upitup:
This PR enables to specify an input field (of type text) to be focused and with the cursor set at the end of the input whenever a default value is specified.

Example usage

group

### DIFF
--- a/src/Form/FieldInput.js
+++ b/src/Form/FieldInput.js
@@ -17,6 +17,20 @@ class FieldInput extends React.Component {
     }
   }
 
+  componentDidMount() {
+    let inputElement = ReactDOM.findDOMNode(this.refs.inputElement);
+    if (inputElement != null &&
+        inputElement.type === 'text' &&
+        this.props.focused) {
+      // Don't interfere with existing transitions
+      setTimeout(function () {
+        var valueLength = inputElement.value.length;
+        inputElement.focus();
+        inputElement.setSelectionRange(valueLength, valueLength);
+      }, 0);
+    }
+  }
+
   bindEvents(attributes) {
     EVENTS.forEach((event) => {
       let htmlEvent = `on${Util.capitalize(event)}`;
@@ -181,6 +195,10 @@ FieldInput.propTypes = {
   // Optional. Which field property is currently being edited
   // (usually passed down from form definition)
   editing: React.PropTypes.string,
+  // Optional. Specify if the field should be focused
+  // Useful in combination with preset values, will set cursor to end of input
+  // (usually passed down from form definition)
+  focused: React.PropTypes.bool,
   // Function to handle when form is submitted
   // (usually passed down from form definition)
   handleSubmit: React.PropTypes.func,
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @rcorral

jfurrow:
This looks great!

src/Form/FieldInput.js
```diff
@@ -181,6 +195,10 @@ FieldInput.propTypes = {
   // Optional. Which field property is currently being edited
   // (usually passed down from form definition)
   editing: React.PropTypes.string,
+  // Optional. Specify if the field should be focused
+  // Useful in combination with preset values, will set cursor to end of input
+  // (usually passed down from form definition)
+  focused: React.PropTypes.bool,
```
rcorral: I say we just default this to true, above on the `defaultProps`.

src/Form/FieldInput.js
```diff
@@ -17,6 +17,20 @@ class FieldInput extends React.Component {
     }
   }
 
+  componentDidMount() {
+    let inputElement = ReactDOM.findDOMNode(this.refs.inputElement);
+    if (inputElement != null &&
+        inputElement.type === 'text' &&
+        this.props.focused) {
+      // Don't interfere with existing transitions
```
rcorral: This won't break transitions that last a long time?

src/Form/FieldInput.js
pierlo-upitup: Are you sure? What if there are multiple input fields? Which one will be focused then (probably the last rendered one)? 
The idea behind this property is to declare **one** field that should be focused when the form is rendered.

src/Form/FieldInput.js
pierlo-upitup: Good shout. `setTimeout` is definitely a hacky workaround. Using `requestAnimationFrame` would be another option, but it still won't solve the reliability problem.

A far better approach here would be to make sure that the transition is finished by binding on the `transitionend` event, and then firing off the `focus()` call. However I am reluctant to let this component do the guesswork to find which of its ancestors it should bind itself onto... 

As we are relying on the (optional) CSS framework being use to stye these components (in our case, Canvas) and not orchestrating these transitions with JS I'm not sure what options we have left. I'm all ears in case you have an 💡 !

mlunoe:
I know I am late to the party, but we can't do with `autoFocus`? That works now. If you add it to the definition it will add focus to your form field.

pierlo-upitup:
@mlunoe what this PR accomplishes is to set the cursor at the end of a pre-defined input field. The `autofocus` attribute won't help us there (at least not consistently).

mlunoe:
Arh! ok, I was missing that detail

src/Form/FieldInput.js
rcorral: Oh good point, didn't think of that.

src/Form/FieldInput.js
rcorral: This is fine for now, but I'm sure we'll run into this problem eventually. So we'll have to fix it then.

PR #259, opened 2016-05-17
Fix update height issue in modal

mlunoe:
This fixes an height update problem with modal. It would not update the height unless you explicitly updated the the container of the component twice.

Before:

After:


mesosphere-web:
Reviewer: @MatApple

jfurrow:
I tested this on the branch where I need this functionality and it works perfectly!

PR #260, opened 2016-05-20
Add field select component

Poltergeist:
This adds a field select component. I am missing tests but it would be great to get already some comments on the component.

image

### DIFF
--- a/src/Form/FieldSelect.js
+++ b/src/Form/FieldSelect.js
@@ -39,7 +39,7 @@ class FieldSelect extends React.Component {
       defaultValue = dropdownItems[0];
     }
 
-    if (typeof defaultValue === 'string'){
+    if (typeof defaultValue === 'string') {
       defaultValue = {
         html: defaultValue,
         id: defaultValue.toLowerCase()
@@ -134,6 +134,7 @@ class FieldSelect extends React.Component {
 
 FieldSelect.defaultProps = {
   columnWidth: 12,
+  handleEvent: function () {},
   formElementClass: 'form-row-element'
 };
 
--- a/src/Form/FieldSelect.js
+++ b/src/Form/FieldSelect.js
@@ -17,14 +17,23 @@ class FieldSelect extends React.Component {
   }
 
   getDropDown(dropdownItems) {
-    let {
-      defaultValue,
-      dropdownMenuButtonClassName,
-      dropdownMenuClassName,
-      dropdownMenuListClassName,
-      dropdownMenuWrapperClassName,
-      persistentID
-    } = this.props;
+    let {defaultValue, persistentID} = this.props;
+
+    let classes = {
+      buttonClassName: 'button dropdown-toggle',
+      dropdownMenuClassName: 'dropdown-menu',
+      dropdownMenuListClassName: 'dropdown-menu-list',
+      wrapperClassName: 'dropdown'
+    };
+
+    classes = Object.keys(classes)
+      .reduce(function (classSet, className) {
+        classSet[className] = classNames(
+          classes[className],
+          this.props[className]
+        );
+        return classSet;
+    }.bind(this), classes);
 
     if (defaultValue == null) {
       defaultValue = dropdownItems[0];
@@ -37,25 +46,13 @@ class FieldSelect extends React.Component {
       };
     }
 
-    let dropdownMenuButtonClassSet = classNames('button dropdown-toggle',
-      dropdownMenuButtonClassName);
-    let dropdownMenuClassSet = classNames('dropdown-menu',
-      dropdownMenuClassName);
-    let dropdownMenuListClassSet = classNames('dropdown-menu-list',
-      dropdownMenuListClassName);
-    let dropdownMenuWrapperClassSet = classNames('dropdown',
-      dropdownMenuWrapperClassName);
-
     return (
-      <Dropdown buttonClassName={dropdownMenuButtonClassSet}
-        dropdownMenuClassName={dropdownMenuClassSet}
-        dropdownMenuListClassName={dropdownMenuListClassSet}
+      <Dropdown {...classes}
+        initialID={defaultValue.id}
         items={dropdownItems}
         onItemSelection={this.handleChange.bind(this)}
-        transition={true}
         persistentID={persistentID}
-        initialID={defaultValue.id}
-        wrapperClassName={dropdownMenuWrapperClassSet} />
+        transition={true} />
     );
   }
 
@@ -190,9 +187,10 @@ FieldSelect.propTypes = {
   formElementClass: React.PropTypes.string,
 
   //Classes for the Dropdown
-  dropdownMenuButtonClassName: React.PropTypes.string,
-  dropdownMenuListClassName: React.PropTypes.string,
+  buttonClassName: React.PropTypes.string,
   dropdownMenuClassName: React.PropTypes.string,
+  dropdownMenuListClassName: React.PropTypes.string,
+  wrapperClassName: React.PropTypes.string
 };
 
 module.exports = FieldSelect;
--- a/src/Form/FieldSelect.js
+++ b/src/Form/FieldSelect.js
@@ -17,8 +17,7 @@ class FieldSelect extends React.Component {
   }
 
   getDropDown(dropdownItems) {
-    let {defaultValue, persistentID} = this.props;
-
+    let {startValue, persistentID} = this.props;
     let classes = {
       buttonClassName: 'button dropdown-toggle',
       dropdownMenuClassName: 'dropdown-menu',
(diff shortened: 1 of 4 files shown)


mesosphere-web:
Reviewer: @jfurrow

src/Form/FieldSelect.js
```diff
@@ -0,0 +1,153 @@
...
+    );
+  }
+
+  render() {
+    let {
+      columnWidth, formGroupClass, formElementClass, formGroupErrorClass,
+      options
+    } = this.props;
+
+    options = this.props.options.map(function (option) {
```
jfurrow: We can just use `options` again instead of `this.props.options`, so `options = options.map(...`

src/Form/FieldSelect.js
```diff
@@ -0,0 +1,153 @@
...
+      if (typeof option === 'string') {
+        return {
+          html: option,
+          id: option.toLowerCase()
+        };
+      }
```
jfurrow: I like this approach!

src/Form/FieldSelect.js
```diff
@@ -0,0 +1,153 @@
+import classNames from 'classnames';
```
jfurrow: I think this is a good opportunity to use `classnames/dedupe`, but perhaps, for performance reasons, others may disagree.

src/Form/FieldSelect.js
```diff
@@ -0,0 +1,153 @@
...
+    let {props} = this;
+    let validationError = props.validationError;
+    return !!(validationError && validationError[props.name]);
+  }
+
+  getDropDown(dropdownItems) {
+    return (
+      <Dropdown buttonClassName="button dropdown-toggle"
+        dropdownMenuClassName="dropdown-menu"
+        dropdownMenuListClassName="dropdown-menu-list"
```
jfurrow: I think we should make these classnames props so we can change them when needed.

jfurrow:
This is looking great! Nice work!

I have a couple of ideas:
* It would be nice to be able to set the `persistentID` prop on the dropdown. When this prop is used, the dropdown always displays that dropdown item. This gives you the opportunity to set some instructional text which will always be displayed.
* Maybe we should set the `initialID` prop as the ID of the first dropdown item, unless it is provided by the user, to avoid the case where there is no text in the dropdown button.

jfurrow:
LGTM once you get tests in there!

Poltergeist:
Added tests

src/Form/FieldSelect.js
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    if (typeof defaultValue === 'string'){
+      defaultValue = {
+        html: defaultValue,
+        id: defaultValue.toLowerCase()
+      };
+    }
+
+    let dropdownMenuButtonClassSet = classNames('button dropdown-toggle',
```
mlunoe: This feels repititive, maybe we could do something like this:
```js
getDropDown(dropdownItems) {
  let {defaultValue, persistentID} = this.props;

  let classes = {
    buttonClassName: {
      defaults: 'button dropdown-toggle',
      propsName: 'dropdownMenuButtonClassName'
    },
    dropdownMenuClassName: {
      defaults: 'dropdown-menu',
      propsName: 'dropdownMenuClassName'
    },
    dropdownMenuListClassName: {
      defaults: 'dropdown-menu-list',
      propsName: 'dropdownMenuListClassName'
    },
    wrapperClassName: {
      defaults: 'dropdown',
      propsName: 'dropdownMenuWrapperClassName'
    }
  };

  classes = Object.keys(classes).reduce(function (classSet, className) {
    let {defaults, propsName} = classes[className];
    classSet[className] = classNames(defaults, this.props[propsName]);
  }, classes);

  return (
    <Dropdown
      items={dropdownItems}
      onItemSelection={this.handleChange.bind(this)}
      transition={true}
      persistentID={persistentID}
      initialID={defaultValue.id}
      {...classes} />
  );
}
```

Or if you keep the names the same you could do something like this:
```js
getDropDown(dropdownItems) {
  let {defaultValue, persistentID} = this.props;

  let classes = {
    buttonClassName: 'button dropdown-toggle',
    dropdownMenuClassName: 'dropdown-menu',
    dropdownMenuListClassName: 'dropdown-menu-list',
    wrapperClassName: 'dropdown'
  };

  classes = Object.keys(classes).reduce(function (classSet, className) {
    classSet[className] = classNames(classes[className], this.props[className]);
  }, classes);

  return (
    <Dropdown
      items={dropdownItems}
      onItemSelection={this.handleChange.bind(this)}
      transition={true}
      persistentID={persistentID}
      initialID={defaultValue.id}
      {...classes} />
  );
}

```

Not sure if that is the best approach, but something that would make this less repetitive would be good

mlunoe: We could maybe even have a utility to do this for classes in general!

Poltergeist: I do like the second approach, thanks for the suggestion will adjust the code to it.

src/Form/FieldSelect.js
```diff
@@ -0,0 +1,196 @@
...
+          this.props[className]
+        );
+        return classSet;
+    }.bind(this), classes);
+
+    if (defaultValue == null) {
+      defaultValue = dropdownItems[0];
+    }
+
+    if (typeof defaultValue === 'string'){
```
rcorral: could you add a space after the closing `)`

src/Form/FieldSelect.js
```diff
@@ -0,0 +1,196 @@
...
+  // If node: returns the node as the label.
+  showLabel: React.PropTypes.oneOfType([
+    React.PropTypes.node,
+    React.PropTypes.string,
+    React.PropTypes.bool
+  ]),
+
+  // Function to handle events like 'change', 'blur', 'focus', etc on the field
+  // (usually passed down from form definition)
+  handleEvent: React.PropTypes.func,
```
rcorral: You don't have this set to required nor do you have a default, and you also don't check to make sure it's a function before invoking. Let's do one of those options to ensure there's no exceptions!

Poltergeist: I added a default since that seems to be the solution on other components.

Poltergeist:
Are there any more action items on this PR?

mlunoe:
When I submit without having selected anything else than what is preselected I get an error:

src/Form/FieldSelect.js
```diff
@@ -0,0 +1,197 @@
+import classNames from 'classnames/dedupe';
+import React from 'react';
+
+import Dropdown from '../Dropdown/Dropdown';
+
+class FieldSelect extends React.Component {
+
+  handleChange(selectedValue) {
+    this.props.handleEvent('change',
+      this.props.name, selectedValue);
```
mlunoe: I find it a little strange that I receive both html and id as data.


Let's just return the `id` here

src/Form/FieldSelect.js
```diff
@@ -0,0 +1,197 @@
...
+
+    let classes = {
+      buttonClassName: 'button dropdown-toggle',
+      dropdownMenuClassName: 'dropdown-menu',
+      dropdownMenuListClassName: 'dropdown-menu-list',
+      wrapperClassName: 'dropdown'
+    };
+
+    classes = Object.keys(classes)
+      .reduce(function (classSet, className) {
```
mlunoe: You can just do this:
```js
    classes = Object.keys(classes).reduce((classSet, className) => {
        classSet[className] = classNames(
          classes[className],
          this.props[className]
        );
        return classSet;
    }, classes);
```

src/Form/FieldSelect.js
```diff
@@ -0,0 +1,197 @@
...
+  // (usually passed down from form definition)
+  handleEvent: React.PropTypes.func,
+
+  // These are the options for the DropDown Component
+  options: React.PropTypes.arrayOf(
+    React.PropTypes.oneOfType([
+      React.PropTypes.string,
+      React.PropTypes.shape({
+        html: React.PropTypes.string,
+        id: React.PropTypes.string
```
mlunoe: 👍

src/Form/__tests__/FieldSelect-test.js
```diff
@@ -0,0 +1,129 @@
...
+var FieldSelect = require('../FieldSelect');
+
+describe('FieldSelect', function () {
+
+  describe('#hasError', function () {
+    it('should return true when error contains name', function () {
+      var instance = instance = TestUtils.renderIntoDocument(
+        <FieldSelect
+          name="username"
+          options={[
```
mlunoe: #nit This could be in one line

src/Form/__tests__/FieldSelect-test.js
```diff
@@ -0,0 +1,129 @@
...
+jest.dontMock('../../Util/KeyboardUtil');
+
+/* eslint-disable no-unused-vars */
+var React = require('react');
+/* eslint-enable no-unused-vars */
+var TestUtils = require('react-addons-test-utils');
+
+var FieldSelect = require('../FieldSelect');
+
+describe('FieldSelect', function () {
```
mlunoe: Let's add a test for this: https://github.com/mesosphere/reactjs-components/pull/260#issuecomment-221135945

mlunoe:
A few small comments, but it looks good @Poltergeist!

src/Form/__tests__/FieldSelect-test.js
mlunoe: I guess this is more in the functionality of the Form component

mlunoe:
🎉

PR #272, opened 2016-05-25
Checkbox radio button follow up

mlunoe:
Aligning checkbox, with how it used to work before.

To adhere to the radio buttons requirements of having multiple fields change at the same time, arrays were passed to the `onChange` handler, and for checkboxes that meant that arrays of singe items was being passed. This, however, doesn't make too much sense for the users perspective, so I changed the handler to handle both cases, i.e. _arrays_ (for radio buttons) and _single items_ (for checkboxes).


mesosphere-web:
Reviewer: @daiyiwhale

daiyiwhale:
✨

PR #276, opened 2016-06-01
Refactor Form model merge into function

kennyt:
This makes it so that only the keys that are different in the definition, are merged into the model

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,3 +20,4 @@
 - #171 – Updated React to 0.14. React is along with other dependencies are now required as peerDependencies.
 - #261 - Added Radio Button component to the Form component.
 - #262 – Change Checkbox component to extend Radio Button component and reuse functionality. Functionality of FieldCheckbox and FieldCheckboxMultiple is now both contained in FieldCheckbox.
+- #276 - Change Form updating to delete fields that are no longer part of the definition.
--- a/src/Form/Form.js
+++ b/src/Form/Form.js
@@ -18,22 +18,22 @@ function findFieldOption(options, field) {
   });
 }
 
-function mergeNewModel(stateModel, currentModelDefinition, nextModelDefinition) {
+function mergeNewModel(stateModel, currentDefinitionModel, nextDefinitionModel) {
   let newModel = {};
   let stateModelClone = Util.extend({}, stateModel);
 
   // This allows us to remove fields on the model, if we remove that field from
   // the definition.
   Object.keys(stateModelClone).forEach(function (modelKey) {
-    if (!nextModelDefinition.hasOwnProperty(modelKey)) {
+    if (!nextDefinitionModel.hasOwnProperty(modelKey)) {
       delete stateModelClone[modelKey];
     }
   });
 
   // This makes sure we only update the keys that are different.
-  Object.keys(nextModelDefinition).forEach(function (key) {
-    if (currentModelDefinition[key] !== nextModelDefinition[key]) {
-      newModel[key] = nextModelDefinition[key];
+  Object.keys(nextDefinitionModel).forEach(function (key) {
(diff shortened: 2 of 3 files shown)


mesosphere-web:
Reviewer: @MatApple

mlunoe:
Could we get a Changelog entry for this?

src/Form/Form.js
```diff
@@ -18,6 +18,18 @@ function findFieldOption(options, field) {
   });
 }
 
+function mergeNewModel(stateModel, currentModelDefinition, nextModelDefinition) {
+  let newModel = {};
+
+  Object.keys(nextModelDefinition).forEach(function (key) {
+    if (currentModelDefinition[key] !== nextModelDefinition[key]) {
```
mlunoe: I think this will be a more solid check:
`if (!currentModelDefinition.hasOwnProperty(key) && nextModelDefinition.hasOwnProperty(key)) {`

However, it depends on the functionality you need here. Do you want it to override if current model is `null` and next is not and when current has a value, say `'value'`, and next is `null`, and when current is empty string and next is not?

src/Form/Form.js
```diff
@@ -63,14 +75,16 @@ class Form extends Util.mixin(BindMixin) {
     let currentModel = this.buildStateObj(props.definition, 'value');
     let nextModel = this.buildStateObj(nextProps.definition, 'value');
     if (!Util.isEqual(currentModel, nextModel)) {
-      nextState.model = Util.extend({}, state.model, nextModel);
+      nextState.model = mergeNewModel(state.model, currentModel, nextModel);
```
mlunoe: Let's test out this functionality, as it is going to impact all usages of `Form`

src/Form/Form.js
kennyt: yes, if the currentModelDefinition value is `null` and it isn't `null` anymore, then we want to replace it. it's not only for when a property is completely new

mlunoe: This means that it will override if the new property is `undefined` or `null` too. As long as that is what we want :)

src/Form/Form.js
```diff
@@ -18,6 +18,28 @@ function findFieldOption(options, field) {
   });
 }
 
+function mergeNewModel(stateModel, currentModelDefinition, nextModelDefinition) {
```
mlunoe: #nit I guess you could just use `currentModel` and `nextModel` here

kennyt: i think i should rename it into `currentDefinitionModel` because `currentModel` is not what it really is. the `stateModel` is the current model

kennyt: done

src/Form/__tests__/Form-test.js
```diff
@@ -75,6 +76,19 @@ describe('Form', function () {
       expect(this.instance.state.erroredFields.username)
         .toEqual('different error');
     });
+
+    it('should update delete field from model if no longer part of definition', function () {
```
jfurrow: Test name here is wrong, should be `should delete field...`

jfurrow:
Minor nit, looks good!

CHANGELOG.md
```diff
@@ -20,3 +20,4 @@
 - #171 – Updated React to 0.14. React is along with other dependencies are now required as peerDependencies.
 - #261 - Added Radio Button component to the Form component.
 - #262 – Change Checkbox component to extend Radio Button component and reuse functionality. Functionality of FieldCheckbox and FieldCheckboxMultiple is now both contained in FieldCheckbox.
+- #276 - Change Form updating to delete fields that are no longer part of the definition.
```
mlunoe: This was added the wrong place. I see some of the ones above. Created a PR for it here: https://github.com/mesosphere/reactjs-components/pull/278

PR #279, opened 2016-06-02
Making form element classes more customizable

mlunoe:


mesosphere-web:
Reviewer: @daiyiwhale

MatApple:
@mlunoe conflicts

mlunoe:
Snap! That `CHANGELOG.md` is going to cause konflix on all PRs

src/Form/FieldCheckbox.js
```diff
@@ -1,4 +1,4 @@
-import classNames from 'classnames';
+import classNames from 'classnames/dedupe';
```
daiyiwhale: with the addition of `classnames/dedupe` should we remove the `classNames` from `defaultProps` and place them in the `classNames()` functions within the component methods?

kennyt: makes sense to me

mlunoe: This inherits from RadioButton and have the classes defined here: https://github.com/mesosphere/reactjs-components/pull/279/files#diff-59f6aa2f62c69bf931c3325ad9a5d1d3L166

src/Form/FieldInput.js
```diff
@@ -86,7 +86,7 @@ class FieldInput extends React.Component {
 
     if (this.hasError()) {
       errorMsg = (
-        <p className={props.helpBlockClass}>
+        <p className={classNames(props.helpBlockClass)}>
```
daiyiwhale: hmmm maybe what you want here is `classNames('default-class default-class-super-short', props.helpBlockClass)`? that way you have your defaults, and the props will either extend or override according to `classnames/dedupe`

src/Form/FieldCheckbox.js
mlunoe: Wait you are right!

src/Form/FieldInput.js
mlunoe: This particular instance gets its overrides and defaults from the form: https://github.com/mesosphere/reactjs-components/pull/279/files#diff-035be6cdc63453ed42b5cf6b6fcf34b8R308.
In the future it would probably be good to have the fields add their own defaults, but that is out of scope for this PR.

daiyiwhale:
Looks good!

PR #288, opened 2016-06-20
Don't make formTag a variable

kennyt:
px

that happens if we make it a variable. it needs to be `this.props.formTag`


mesosphere-web:
Reviewer: @rcorral

rcorral:
Seems legit

PR #290, opened 2016-06-21
Don`t transform field select ids

### LINKED ISSUE #289: FieldSelect should not transform option ids (values)
The `FieldSelect` component should not transform option ids (values), as this may result in issues.

orlandohohmeier:
Remove the FieldSelect options id to lower case calls.

Closes #289


mesosphere-web:
Reviewer: @daiyiwhale

PR #292, opened 2016-06-22
Find existing ID in dropdown options

MatApple:
If we provide a string as `startValue` then we want it to match an existing option first (if one is available).


mesosphere-web:
Reviewer: @kennyt

PR #294, opened 2016-06-27
Properly obey initial sort order

jfurrow:
Fixes bug where the `Table` would not obey its initial sort order.

### DIFF
--- a/src/Table/Table.js
+++ b/src/Table/Table.js
@@ -57,8 +57,11 @@ class Table extends React.Component {
   }
 
   componentWillMount() {
-    if (this.props.sortBy.prop) {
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @MatApple

src/Table/Table.js
```diff
@@ -57,8 +57,11 @@ class Table extends React.Component {
   }
 
   componentWillMount() {
-    if (this.props.sortBy.prop) {
-      this.handleSort(this.props.sortBy.prop);
+    if (!!this.props.sortBy) {
```
orlandohohmeier: `this.props.sortBy != null`

jfurrow: Done!

PR #296, opened 2016-06-27
Move padding to inside of scroll container

jfurrow:
This PR moves the padding for the modal's content to inside the scroll container.

Here it is in action:


mesosphere-web:
Reviewer: @rcorral

PR #301, opened 2016-06-30
Form fields only update if props change

kennyt:
This improves performance in huge forms by a large degree

### DIFF
--- a/src/Form/FieldCheckbox.js
+++ b/src/Form/FieldCheckbox.js
@@ -7,10 +7,6 @@ import IconCheckbox from './icons/IconCheckbox';
 import Util from '../Util/Util';
 
 class FieldCheckbox extends FieldRadioButton {
-  shouldComponentUpdate(nextProps) {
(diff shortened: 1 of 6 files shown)


mesosphere-web:
Reviewer: @MatApple

src/Form/FieldCheckbox.js
```diff
@@ -7,6 +7,9 @@ import IconCheckbox from './icons/IconCheckbox';
 import Util from '../Util/Util';
 
 class FieldCheckbox extends FieldRadioButton {
+  shouldComponentUpdate(nextProps) {
```
mlunoe: This extends FieldRadioButton, so if that implements it, this doesn't have to

mlunoe:
One comment and then :shipit: !

src/Form/FieldCheckbox.js
kennyt: done

PR #302, opened 2016-07-07
Add release-beta script

kennyt:
Currently, I have the `npm publish` in the `postversion` script. the `postversion` automatically tags the publish with the version, which means it won't be on `latest`. the functionality will have to change if we ever want to create a `release-patch` script

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,9 @@
     "livereload": "NODE_ENV='development' ./node_modules/.bin/gulp docs:livereload",
     "npm-publish": "npm publish ./",
     "test": "jest",
-    "release": "npm run clean && npm run test && ./node_modules/.bin/gulp eslint && npm run dist-src && npm publish ./"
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @mlunoe

mlunoe:
Could we get some documentation on how to use this? There is a tagging and publish section: https://github.com/mesosphere/reactjs-components/blob/master/CONTRIBUTING.md#creating-a-tag-and-publishing

Which set instructions would you go through to run this? `npm run release-beta`?
What does `npm version prerelease` do?

That will also run the postversion script which will also publish it, right?
But that means it publishes before you have created and merged the PR with the version. Is that intentional?
I think I need a little bit more context here

mlunoe:
@MatApple Hmm... I'm still not sure how this works. Could you elaborate please?

PR #303, opened 2016-07-08
Add useExternalErrors prop to Form

kennyt:
Before, the form would get errors from definition + it'll keep it's own state from user interaction for errors. This props make it so that errors are completely driven from the form definition.

### DIFF
--- a/src/Form/Form.js
+++ b/src/Form/Form.js
@@ -88,14 +88,18 @@ class Form extends Util.mixin(BindMixin) {
       nextState.model = mergeNewModel(state.model, currentModel, nextModel);
     }
 
-    let currentErrors = this.buildStateObj(props.definition, 'showError');
-    let nextErrors = this.buildStateObj(nextProps.definition, 'showError');
-    if (!Util.isEqual(currentErrors, nextErrors)) {
-      nextState.erroredFields = mergeNewModel(
-        state.erroredFields,
-        currentErrors,
-        nextErrors
-      );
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @jfurrow

src/Form/Form.js
```diff
@@ -88,14 +88,18 @@ class Form extends Util.mixin(BindMixin) {
...
 
-    let currentErrors = this.buildStateObj(props.definition, 'showError');
-    let nextErrors = this.buildStateObj(nextProps.definition, 'showError');
-    if (!Util.isEqual(currentErrors, nextErrors)) {
-      nextState.erroredFields = mergeNewModel(
-        state.erroredFields,
-        currentErrors,
-        nextErrors
-      );
+    if (this.props.useExternalErrors) {
```
orlandohohmeier: Do we really need this prop? Shouldn't it be enough to always render the errors from the definition if defined?

kennyt: yeah we do need this prop. for example, let's say you pass a definition to a form with no errors set. then through user interaction and the built-in form validation, it creates some errors. if we always pull errors from definition, the next time we render, it'll blow away that internal error.

a solution to this would be to redo the form so that it's stateless and just passes events back up to the parent and then we can fully just pull errors from the definition. i like that idea, but it's going to be too big of a change for now

orlandohohmeier: Alright, thanks for the explanation.

PR #305, opened 2016-07-26
Use fixed positioning for Dropdown

jfurrow:
This PR changes the dropdown to use fixed positioning, which allows it to render outside of elements whose overflowing content is clipped.

See it in action:

It will close automatically when its parent container scrolls.

One breaking change was introduced with this refactor. To anchor dropdowns to the right of the dropdown button, we must now do it via a boolean passed to the `Dropdown` called `anchorRight`. This is due to the fixed positioning.


mesosphere-web:
Reviewer: @mlunoe

PR #308, opened 2016-07-27
Find native scroll container if Gemini isn't needed

jfurrow:
This PR introduces the ability to provide a selector via the prop `scrollContainerParentSelector`. If it matches the parent of the `scrollContainer` element, the scroll event handler will be attached to the parent element rather than `scrollContainer`.

This is necessary due to Gemini changing which element handles overflowing content dependent upon OS-specific settings. If the OS does not render overlay scrollbars, Gemini handles scrolling in a div with class `gm-scroll-view`. If the OS does render native overlay scrollbars, Gemini does nothing except add DOM nodes, so the direct parent of the Gemini component is left responsible for the scroll behavior.

This PR also exposes the `matchesFn` in `DOMUtil` which is now consumed by the `Dropdown`.

### DIFF
--- a/src/Dropdown/Dropdown.js
+++ b/src/Dropdown/Dropdown.js
@@ -185,16 +185,20 @@ class Dropdown extends Util.mixin(BindMixin, KeyDownMixin) {
     let {scrollContainer, scrollContainerParentSelector} = this.props;
 
     if (typeof scrollContainer === 'string') {
-      let scrollContainerSelector = null;
+      // Find the closest scrolling element by the specified selector.
+      scrollContainer = DOMUtil.closest(ReactDOM.findDOMNode(this),
+        scrollContainer) || window;
 
-      if (scrollContainerParentSelector == null) {
-        scrollContainerSelector = scrollContainer;
-      } else {
-        scrollContainerSelector = scrollContainerParentSelector;
+      let {parentElement} = scrollContainer;
+
+      // If the user specified scrollContainerParentSelector, we check to see
+      // if the parent scrolling element matches the specified parent selector.
+      if (scrollContainer != window && scrollContainerParentSelector != null
+        && parentElement != null && parentElement[DOMUtil.matchesFn](
+          scrollContainerParentSelector
+        )) {
+        scrollContainer = parentElement;
       }
-
-      return DOMUtil.closest(ReactDOM.findDOMNode(this),
-        scrollContainerSelector) || window;
     }
 
     return scrollContainer;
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @MatApple

src/Dropdown/Dropdown.js
```diff
@@ -47,6 +47,25 @@ class Dropdown extends Util.mixin(BindMixin, KeyDownMixin) {
...
   }
 
+  componentDidMount() {
+    if (typeof this.props.scrollContainer === 'string') {
+      this.container = DOMUtil.closest(ReactDOM.findDOMNode(this),
+        this.props.scrollContainer) || window;
+
+      if (this.container.parentElement != null
+        && this.props.scrollContainerParentSelector != null
+        && this.container.parentElement[DOMUtil.matchesFn](
```
MatApple: Will it always be the immediate parent? Or should we walk up the tree till we find a match?

jfurrow: Yeah, good call!

jfurrow: Alright this is done, and I think I made it a bit more readable, too!

src/Dropdown/Dropdown.js
```diff
@@ -47,6 +47,25 @@ class Dropdown extends Util.mixin(BindMixin, KeyDownMixin) {
     }
   }
 
+  componentDidMount() {
```
mlunoe: Coudl we do something like this instead:
```js
getContainer() {
    let {scrollContainer, scrollContainerParentSelector} = this.props;
    if (typeof scrollContainer === 'string') {
      let container = DOMUtil.closest(
        ReactDOM.findDOMNode(this),
        scrollContainer
      ) || window;

      if (container.parentElement != null
        && scrollContainerParentSelector != null
        && container.parentElement[DOMUtil.matchesFn](
          scrollContainerParentSelector
        )) {
        return container.parentElement;
      }

      return container;
    }

    return scrollContainer;
  }

 componentDidMount() {
    this.container = this.getContainer();
  }
```

src/Util/DOMUtil.js
```diff
@@ -120,7 +104,23 @@ const DOMUtil = {
     } else {
       return element.scrollTop;
     }
-  }
+  },
+
+  matchesFn: (function () {
```
mlunoe: Is there a particular reason why this needs to be on scope now?

src/Dropdown/Dropdown.js
mlunoe: Updated the comment

src/Util/DOMUtil.js
jfurrow: Nope. Removed!

src/Dropdown/Dropdown.js
MatApple: 😎

mlunoe:
Looks great! Works great! :shipit:

PR #311, opened 2016-08-01
Fix beta release script

jfurrow:
This PR prepends the text `version-` to the version tag, because NPM rejects any tag that begins with `v` or a number. It also fixes the `CONTRIBUTING.md` document to include the usage of the script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "test": "jest",
     "release": "npm run clean && npm run test && ./node_modules/.bin/gulp eslint && npm run dist-src && npm publish ./",
     "release-beta": "npm run clean && npm run test && ./node_modules/.bin/gulp eslint && npm run dist-src && npm version prerelease",
-    "postversion": "npm publish ./ --tag $npm_package_version"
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

package.json
```diff
@@ -81,7 +81,7 @@
     "test": "jest",
     "release": "npm run clean && npm run test && ./node_modules/.bin/gulp eslint && npm run dist-src && npm publish ./",
     "release-beta": "npm run clean && npm run test && ./node_modules/.bin/gulp eslint && npm run dist-src && npm version prerelease",
-    "postversion": "npm publish ./ --tag $npm_package_version"
+    "postversion": "npm publish ./ --tag version-$npm_package_version"
```
mlunoe: Why do we prepend `version-`? we have never done that before. Usually we have done the `-beta.xx` part

jfurrow: Because this is the full version — e.g. `0.14.0.beta-19`. So my prefix will make it `version-0.14.0.beta-19`

PR #314, opened 2016-08-02
Implement auto-expanding "TextArea"

jfurrow:
This PR replaces all `textarea` elements with `div` elements with `contenteditable="true"`.

It seems to work inside DCOS UI, but please test this heavily.

### DIFF
--- a/src/Form/FieldInput.js
+++ b/src/Form/FieldInput.js
@@ -2,13 +2,14 @@ import classNames from 'classnames/dedupe';
 import React from 'react';
 import ReactDOM from 'react-dom';
 
+import BindMixin from '../Mixin/BindMixin';
 import IconEdit from './icons/IconEdit';
 import KeyboardUtil from '../Util/KeyboardUtil';
 import Util from '../Util/Util';
 
 const EVENTS = ['blur', 'change', 'focus'];
 
-class FieldInput extends React.Component {
+class FieldInput extends Util.mixin(BindMixin) {
   shouldComponentUpdate(nextProps, nextState) {
     return !Util.isEqual(this.props, nextProps) || !Util.isEqual(this.state, nextState);
   }
--- a/src/Form/FieldInput.js
+++ b/src/Form/FieldInput.js
@@ -9,8 +9,8 @@ import Util from '../Util/Util';
 const EVENTS = ['blur', 'change', 'focus'];
 
 class FieldInput extends React.Component {
-  shouldComponentUpdate(nextProps) {
-    return !Util.isEqual(this.props, nextProps);
+  shouldComponentUpdate(nextProps, nextState) {
+    return !Util.isEqual(this.props, nextProps) || !Util.isEqual(this.state, nextState);
   }
 
   componentDidUpdate() {
--- a/src/Form/FieldInput.js
+++ b/src/Form/FieldInput.js
@@ -10,7 +10,7 @@ import Util from '../Util/Util';
 const EVENTS = ['blur', 'change', 'focus'];
 
 class FieldInput extends Util.mixin(BindMixin) {
-  shouldComponentUpdate(nextProps, nextState) {
+  shouldComponentUpdate(nextProps) {
     return !Util.isEqual(this.props, nextProps);
   }
 
--- a/src/Form/FieldInput.js
+++ b/src/Form/FieldInput.js
@@ -11,7 +11,7 @@ const EVENTS = ['blur', 'change', 'focus'];
 
 class FieldInput extends Util.mixin(BindMixin) {
   shouldComponentUpdate(nextProps, nextState) {
-    return !Util.isEqual(this.props, nextProps) || !Util.isEqual(this.state, nextState);
+    return !Util.isEqual(this.props, nextProps);
   }
 
   componentDidUpdate() {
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @MatApple

src/Form/FieldInput.js
```diff
@@ -9,8 +9,8 @@ import Util from '../Util/Util';
 const EVENTS = ['blur', 'change', 'focus'];
 
 class FieldInput extends React.Component {
-  shouldComponentUpdate(nextProps) {
-    return !Util.isEqual(this.props, nextProps);
+  shouldComponentUpdate(nextProps, nextState) {
+    return !Util.isEqual(this.props, nextProps) || !Util.isEqual(this.state, nextState);
```
jfurrow: `FieldTextarea` is the first component extending `FieldInput` that makes use of `setState()`, so we need to check that here as well.

src/Form/FieldTextarea.js
```diff
@@ -5,30 +5,94 @@ import React from 'react';
...
+    super(...arguments);
+
+    this.state = {height: 0};
+
+    this.updateTextareaHeight = Util.throttle(this.updateTextareaHeight, 100);
+    this.handleContentEditableBlur = this.handleContentEditableBlur.bind(this);
+    this.handleContentEditableChange =
+      this.handleContentEditableChange.bind(this);
+    this.handleContentEditableFocus =
+      this.handleContentEditableFocus.bind(this);
```
jfurrow: Since we are extending `FieldInput`, we can't use the `Bind` mixin.

src/Form/FieldTextarea.js
```diff
@@ -5,30 +5,94 @@ import React from 'react';
...
+  handleContentEditableChange(event) {
+    let {props} = this;
+
+    this.updateTextareaHeight(event.target);
+    props.handleEvent('change', props.name, event.target.innerText, event);
+  }
+
+  handleContentEditableFocus(event) {
+    let {props} = this;
+    props.handleEvent('focus', props.name, event.target.innerText, event);
```
jfurrow: We have to "proxy" `handleEvent` because `FieldInput#handleEvent` looks at `event.target.value`, which is `undefined` for `contentEditable` elements. Instead we provide `innerText` which contains text (with line breaks and other whitespace characters).

src/Form/FieldInput.js
rcorral: Let's override `shouldComponentUpdate` inside of `FieldTextarea` and not call super.

src/Form/FieldTextarea.js
rcorral: What if we have `FieldInput` implement the BindMixin?
It's [smart](https://github.com/mesosphere/reactjs-components/blob/master/src/Mixin/BindMixin.js#L3) enough to not try to bind things if methodsToBind isn't defined.

src/Form/FieldTextarea.js
```diff
@@ -5,30 +5,94 @@ import React from 'react';
...
+    this.updateTextareaHeight = Util.throttle(this.updateTextareaHeight, 100);
+    this.handleContentEditableBlur = this.handleContentEditableBlur.bind(this);
+    this.handleContentEditableChange =
+      this.handleContentEditableChange.bind(this);
+    this.handleContentEditableFocus =
+      this.handleContentEditableFocus.bind(this);
+  }
+
+  componentWillMount() {
+    this.setState({height: this.props.minHeight});
```
rcorral: Why don't we just set this default inside of the constructor?

src/Form/FieldInput.js
jfurrow: I think this belongs here, it doesn't make sense for these components not to update when their state changes.

src/Form/FieldTextarea.js
jfurrow: Done! I was thinking props aren't available in the constructor. But they are....

src/Form/FieldInput.js
rcorral: But the check if state has changed is only useful in the `FieldTextarea` right?
It's not important for password right?

src/Form/FieldTextarea.js
```diff
@@ -5,30 +5,92 @@ import React from 'react';
...
 import IconEdit from './icons/IconEdit';
+import Util from '../Util/Util';
+
+class FieldTextarea extends FieldInput {
+  constructor() {
+    super(...arguments);
+
+    this.state = {height: this.props.minHeight};
+
+    console.log(this.props);
```
rcorral: heya

src/Form/FieldTextarea.js
jfurrow: Done!

src/Form/FieldInput.js
jfurrow: Correct, at this point in time it is only useful for `FieldTextarea`.

src/Form/FieldTextarea.js
jfurrow: haha damnit, you caught it! This is removed in a future commit.

jfurrow:
I just realized the sue of `contenteditable

src/Form/FieldInput.js
jfurrow: Alright I moved the equality check to just `FieldTextarea`

PR #316, opened 2016-08-02
Add support for dropdown menu matching button width

jfurrow:
This PR sets the `left` and `right` properties of the dropdown menu to match the `left` and `right` position of the dropdown button.

PR #317, opened 2016-08-02
Pass matchButtonWidth to Dropdown

jfurrow:
This will only make a difference when #316 merges.

### DIFF
--- a/src/Form/FieldSelect.js
+++ b/src/Form/FieldSelect.js
@@ -26,7 +26,7 @@ class FieldSelect extends React.Component {
   }
 
   getDropDown(dropdownItems) {
-    let {startValue, persistentID} = this.props;
+    let {matchButtonWidth, startValue, persistentID} = this.props;
     let classes = {
       buttonClassName: 'button dropdown-toggle',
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

src/Form/FieldSelect.js
```diff
@@ -73,6 +73,7 @@ class FieldSelect extends React.Component {
       <Dropdown {...classes}
         initialID={startValue.id}
         items={dropdownItems}
+        matchButtonWidth={matchButtonWidth}
```
mlunoe: Shouldn't we be more dynamic in the sense where we `omit` props that we will use here and pass down the rest of the props dynamically like so `{...restOfProps}`?

jfurrow: Yeah, let's do it for this case.

jfurrow: This is done now. We just need to be sure not to do this for actual DOM elements, because React 15 will throw a warning. https://facebook.github.io/react/warnings/unknown-prop.html

PR #318, opened 2016-08-03
[CNVS-Upgrade] Update table styles

jfurrow:
Adds a `virtual-list` class when it's rendered. This will allow us to proper apply border styles.


jfurrow:
Closing to open against a feature branch 😞

PR #320, opened 2016-08-04
Collapsible proptype blocks

rcorral:
gif

Looking for feedback on looks of this. 
ping: @jfurrow


mesosphere-web:
Reviewer: @mlunoe

jfurrow:
I love the idea of the collapsable blocks! What do you think about making the actual example JSX collapsed, too? That would be really helpful to let people see the actual components easily. This is a pain point for me when I'm working on developing a new component.

Regarding the gradient, I think it might look nicer if the gradient were "inside" the container, like so:

mlunoe:
Nice suggestion @jfurrow.
This is cool @rcorral!

rcorral:
Applied updated gradients
img

jfurrow:
@rcorral I pushed two commits, feel free to merge if you're cool with them!

PR #323, opened 2016-08-04
Fix form prop definitions

mlunoe:
Make sure to be more liberal in the values accepted by different fields in the form component


mesosphere-web:
Reviewer: @jfurrow

PR #335, opened 2016-08-24
Add help block to input field

mlunoe:
This adds the ability to have help blocks in addition to error messages

### DIFF
--- a/docs/src/Form/index.js
+++ b/docs/src/Form/index.js
@@ -28,7 +28,7 @@ class FormExample extends React.Component {
       {
         fieldType: 'password',
         name: 'Password',
-        helpBlock: 'If you want to make your password strong you could use a password manager',
+        helpBlock: 'Setting helpBlock can be used to display helpful text',
         required: true,
         showLabel: true,
         validation: function (value) {
--- a/docs/src/Form/index.js
+++ b/docs/src/Form/index.js
@@ -214,7 +214,7 @@ class FormExample extends React.Component {
       {
         fieldType: 'password',
         name: 'Password',
-        helpBlock: 'If you want to make your password strong you could use a password manager',
+        helpBlock: 'Setting helpBlock can be used to display helpful text',
         required: true,
         showLabel: true,
         validation: function (value) {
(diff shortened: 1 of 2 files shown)


src/Form/FieldInput.js
```diff
@@ -222,6 +233,8 @@ FieldInput.propTypes = {
   handleEvent: React.PropTypes.func,
   // Optional label to add
   label: React.PropTypes.string,
+  // Optional hepl block
+  helpBlock: React.PropTypes.node,
```
jfurrow: Let's add `helpBlockClass` to `propTypes`

mlunoe: It's already there

jfurrow: Ahh cool, I didn't realize that. Should we call `classnames` on it as we do in [here](https://github.com/mesosphere/reactjs-components/blob/dfea31110f87fe945419749fefde3fa84e2f08c7/src/Form/FieldInput.js#L94)?

mlunoe: Arh, yes we should. Done

mlunoe:
ping @jfurrow @philipnrmn

jfurrow:
:ship:

src/Form/FieldInput.js
```diff
@@ -222,6 +237,8 @@ FieldInput.propTypes = {
   handleEvent: React.PropTypes.func,
   // Optional label to add
   label: React.PropTypes.string,
+  // Optional hepl block
```
philipnrmn: s/hepl/help

docs/src/Form/index.js
```diff
@@ -28,6 +28,7 @@ class FormExample extends React.Component {
       {
         fieldType: 'password',
         name: 'Password',
+        helpBlock: 'If you want to make your password strong you could use a password manager',
```
philipnrmn: Not a fan of this wording - the tone is at odds with what we're using elsewhere.

philipnrmn:
No issue with the code, but I'm not sure I like the actual warning much! Happy to merge if we commit to reviewing it soon though.

src/Form/FieldInput.js
mlunoe: Done

docs/src/Form/index.js
mlunoe: Changed it up a bit. Had a conversation offline and came up with something that we could both agree on

mlunoe:
Fixed

PR #350, opened 2016-09-15
Remove more styling from content-editable children

jfurrow:
This PR removes additional styling from children of `.content-editable` to prevent things like the following:


rcorral:
merge after tests pass

PR #354, opened 2016-09-21
[CNVS Upgrade] Fix modal animations

jfurrow:
This PR fixes the animations of modals. Previously they weren't animating on entrance due to the way the transition property was being applied.

It also adjusts the default transition duration to 300ms which matches the transition duration in the CSS.


mlunoe:
Looks like the animation is not working properly the first time it enters:

mlunoe:
This was not ready to merge

PR #356, opened 2016-09-22
[CNVS Upgrade] Allow custom close button

jfurrow:
This PR allows users of the `Modal` to pass an element to use for a close button.

Previously there was a boolean property, `showCloseButton`, which rendered a specific set of nodes. Now, there is a `closeButton` prop which will cause a render of a `button` element with the provided node as its child.

### DIFF
--- a/src/Modal/ModalContents.js
+++ b/src/Modal/ModalContents.js
@@ -218,20 +218,16 @@ class ModalContents extends Util.mixin(BindMixin, KeyDownMixin) {
 
   getCloseButton() {
     let props = this.props;
-    if (!props.showCloseButton) {
-      return null;
+
+    if (props.closeButton) {
+      return (
+        <button className={props.closeButtonClass} onClick={this.closeModal}>
+          {props.closeButton}
+        </button>
+      );
     }
 
-    return (
-      <a
-        className={props.closeButtonClass}
-        onClick={this.closeModal}>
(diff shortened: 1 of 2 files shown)


philipnrmn:
The code LGTM. Will you update the modals in dcos-ui to match this?

jfurrow:
@philipnrmn Yep! There are three modals in DC/OS UI which use the `showCloseButton` prop so I'll ensure that they're handled properly!

src/Modal/ModalContents.js
```diff
@@ -218,20 +218,16 @@ class ModalContents extends Util.mixin(BindMixin, KeyDownMixin) {
 
   getCloseButton() {
     let props = this.props;
-    if (!props.showCloseButton) {
-      return null;
+
+    if (props.closeButton) {
+      return (
+        <button className={props.closeButtonClass} onClick={this.closeModal}>
```
mlunoe: Talked with @jfurrow offline, and decided to make this completely free of boundaries, so it doesn't even wrap it in a button. That way you could just pass:
```js
closeButton={<button onClick={() => { this.setState({open: false}); }}>Close</button>}
```

jfurrow: Done!

mlunoe:
Looks good! Feel free to ⛵ once tests pass

PR #359, opened 2016-09-26
[CNVS Upgrade] Refactor modal height logic

jfurrow:
This PR aims to simplify the height calculation logic of the `ModalContents` component.

Highlights:
* We use `setState` instead of `forceUpdate` to re-render the modal.
* We calculate the height of the content only when absolutely necessary.
* We make use of `requestAnimationFrame` to help reduce jank.
* The maximum height of the modal is determined by CSS rather than with the `maxHeightPercentage` prop.
* The props API of this component was reduced significantly.
* When `useGemini` is set to `false` (which is now _default_), there is no height calculation done at all.

### DIFF
--- a/src/Modal/ModalContents.js
+++ b/src/Modal/ModalContents.js
@@ -90,28 +90,31 @@ class ModalContents extends Util.mixin(BindMixin, KeyDownMixin) {
   handleWindowResize() {
     let {props, state} = this;
 
-    // Recalculate the height of the modal if it's open and using Gemini.
-    if (props.open && props.useGemini) {
-      let viewportHeight = DOMUtil.getViewportHeight();
-
-      // If the height of the viewport is getting shorter, or if it's growing
-      // while the height is currently constrained, then we reset the restrained
-      // height to null which will cause the height to be recalculated on the
-      // next render.
-      if (viewportHeight < this.lastViewportHeight
-        || (viewportHeight > this.lastViewportHeight
-          && state.height !== null)) {
-        this.setState({height: null});
-      }
-
-      this.lastViewportHeight = viewportHeight;
+    // Return early if the modal is closed or not using Gemini.
+    if (!props.open || !props.useGemini) {
+      return;
     }
+
+    let viewportHeight = DOMUtil.getViewportHeight();
+
+    // If the height of the viewport is getting shorter, or if it's growing
+    // while the height is currently constrained, then we reset the restrained
+    // height to null which will cause the height to be recalculated on the
+    // next render.
+    if (viewportHeight < this.lastViewportHeight
+      || (viewportHeight > this.lastViewportHeight
+        && state.height !== null)) {
+      this.setState({height: null});
+    }
+
+    this.lastViewportHeight = viewportHeight;
   }
 
   calculateContentHeight() {
-    let innerContentHeight = this.refs.innerContent
-      .getBoundingClientRect().height;
-    let innerContentContainerHeight = this.refs.innerContentContainer
+    let {innerContent, innerContentContainer} = this.refs;
+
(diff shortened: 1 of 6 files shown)


src/Modal/ModalContents.js
```diff
@@ -253,34 +154,34 @@ class ModalContents extends Util.mixin(BindMixin, KeyDownMixin) {
...
   }
 
-  getModalContent(useScrollbar, contentHeight) {
-    if (!useScrollbar) {
-      return (
-        <div className={this.props.scrollContainerClass}>
-          {this.props.children}
-        </div>
-      );
+  getModalContent(useGemini) {
```
MatApple: Don't need to pass the prop!

jfurrow: Done!

MatApple:
LGTM - nice work @jfurrow!

jfurrow:
Thanks @MatApple!

I just updated this to make use of `jest/setupEnv.js` to monkeypatch `global.requestAnimationFrame` instead of doing it in the `ModalContents` test itself.

src/Modal/ModalContents.js
```diff
@@ -80,144 +88,43 @@ class ModalContents extends Util.mixin(BindMixin, KeyDownMixin) {
...
-      if (maxHeightDifference > 0 &&
-        (heightInfo.contentHeight + difference < innerContentHeight)) {
-        // Changes this.heightInfo as this is a reference
-        heightInfo.height += maxHeightDifference;
-        heightInfo.contentHeight += maxHeightDifference;
-        update = true;
+    let {props, state} = this;
+
+    // Recalculate the height of the modal if it's open and using Gemini.
+    if (props.open && props.useGemini) {
```
rcorral: I would prefer to see us return early.
```
if (!props.open || !props.useGemini) {
  return;
}
```

I like pyramids, just not in code.

src/Modal/ModalContents.js
```diff
@@ -80,144 +88,43 @@ class ModalContents extends Util.mixin(BindMixin, KeyDownMixin) {
...
+      let viewportHeight = DOMUtil.getViewportHeight();
+
+      // If the height of the viewport is getting shorter, or if it's growing
+      // while the height is currently constrained, then we reset the restrained
+      // height to null which will cause the height to be recalculated on the
+      // next render.
+      if (viewportHeight < this.lastViewportHeight
+        || (viewportHeight > this.lastViewportHeight
+          && state.height !== null)) {
+        this.setState({height: null});
```
rcorral: Will it be OK if this `setState becomes async? (just looking at the line below)
Should we do this instead?
```
this.setState({height: null}, () => {
  this.lastViewportHeight = viewportHeight;
});
```

src/Modal/ModalContents.js
```diff
@@ -80,144 +88,43 @@ class ModalContents extends Util.mixin(BindMixin, KeyDownMixin) {
...
   }
 
-  checkHeight() {
-    let prevHeightInfo = this.heightInfo;
-
-    // Calculate height and call a render on first render cycle
-    this.heightInfo = this.calculateModalHeight();
-    this.checkContentHeightChange(prevHeightInfo, this.heightInfo);
+  calculateContentHeight() {
+    let innerContentHeight = this.refs.innerContent
```
rcorral: This could read a little better if you did:
```
let {innerContent, innerContentContainer} = this.refs;
...
```

jfurrow:
@rcorral The demo is and extra props removed from `Confirm` modal!

src/Modal/ModalContents.js
jfurrow: Done! lol @ pyramids

PR #372, opened 2016-11-16
Making modals stack on top of each other when multiple are open

mlunoe:
This PR corrects a bug, where modals would not stack correctly.

Before: | After:
------------ | -------------
 |

### DIFF
--- a/src/Modal/ModalContents.js
+++ b/src/Modal/ModalContents.js
@@ -1,3 +1,4 @@
+import classNames from 'classnames/dedupe';
 import GeminiScrollbar from 'react-gemini-scrollbar';
 /* eslint-disable no-unused-vars */
 import React, {PropTypes} from 'react';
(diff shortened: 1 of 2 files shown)


jfurrow:
The real magic in this PR is the table in your description... TIL

src/Modal/ModalContents.js
```diff
@@ -286,6 +286,7 @@ ModalContents.defaultProps = {
   footerClass: 'modal-footer',
   headerClass: 'modal-header',
   modalClass: 'modal modal-large',
+  modalWrapperClass: 'modal-wrapper',
```
jfurrow: Just a heads up, we do make use of `modalWrapperClass` in `dcos-ui`. So maybe in this case it would make sense to add `modal-wrapper` as an inline string when creating the classes for this element?

src/Modal/styles.less
```diff
@@ -4,12 +4,17 @@
 // * @import (inline) '../../react-gemini-scrollbar/node_modules/gemini-scrollbar/gemini-scrollbar.css';
 
 & when (@modal-enabled) {
+  .modal-wrapper {
```
jfurrow: I'm surprised this wasn't caught by the linter, but could we put a newline above `.modal-wrapper`?

src/Modal/ModalContents.js
mlunoe: Done!

src/Modal/styles.less
mlunoe: done!

jfurrow:
LGTM! :ship:

PR #379, opened 2016-12-12
Update CNVS to 1.1.14

jfurrow:


package.json
```diff
@@ -67,7 +67,7 @@
     "webpack": "1.12.13"
   },
   "peerDependencies": {
-    "cnvs": "1.1.3",
+    "cnvs": "~1.1.4",
```
rcorral: Do we want the minor/patch matching too since we weren't doing it before?

mlunoe: Yeah, I think we should keep the tilde, so it will not break `dcos-ui` as soon as we are out of sync with `reactjs-components`

jfurrow: @mlunoe and I spoke offline about this and decided to go with `~1.1.x` for the `peerDependency`.

PR #380, opened 2016-12-17
Replace KeyDownMixin (refs DCOS-11504)

jfurrow:
This PR removes the `KeyDownMixin` and just implements the event handlers in the components that need them.

The `KeyDownMixin` was a little too generic for our use case. It add keydown listeners for each instance of the `Dropdown` and `Modal` component, when it really only needs to be added when the components are visible.

This caused a problem with our modal where the browser would drop the animation when the `esc` key was pressed.

You can see the difference of the execution times in the following screenshots. In both cases, the timeline was recorded at the moment the `esc` key was pressed in the Service Create Modal in DC/OS.

Before (no animation occurred):

After (the animation occurs after the single handler is called):


mlunoe:
Merging as this has been outstanding for a while

PR #381, opened 2016-12-26
Updated CNVS from 1.1.3 to 1.1.6

ashenden:
Notable updates:
* Updated .button-flush with directional modifiers
* Fixed disabled buttons of type *-link
* Adjusted background of .form-control and .form-control-toggle
* Updated background style of .form-control style variants
* URI encode inline SVG strings
* Updated styling of `disabled` state for `form-control` component

Full diff here for reviewers: https://github.com/mesosphere/cnvs/compare/4e301e8951b74b7c4d6f959ef255651f7da6a11d...d2eb4d1477a0aab827071db210a7c638f9b3cc45

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "babel-preset-react": "6.5.0",
     "babel-runtime": "6.5.0",
     "browser-sync": "2.11.1",
-    "cnvs": "1.1.4",
+    "cnvs": "1.1.5",
     "envify": "3.4.0",
     "eslint": "2.2.0",
     "eslint-plugin-react": "4.0.0",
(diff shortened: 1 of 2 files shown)


mlunoe:
@ashenden Could you add details about what this CNVS update includes?
1.1.3 -> 1.1.6

ashenden:
@mlunoe comments added in description.

mlunoe:
Added full diff to make it easier to review

mlunoe:
Put up a fix for the form bug in the docs here: https://github.com/mesosphere/reactjs-components/pull/382

mlunoe:
Fixed the classes for ya here:

mlunoe:
We talked about offline how the help-text turning red is a little weird and @ashenden agreed that we need to find a solution that works for help-text and error text (since it is now just one, i.e. feedback).

It is not a breaking functionality so this will work for now, but design team will talk it through and come up with a solution to update with at a later stage.

PR #383, opened 2016-12-27
Scroll event not being bubbled in virtual list, ref. DCOS-11781

mlunoe:
---
Same fix as here: https://github.com/mesosphere/reactjs-components/pull/384

---

This PR fixes a bug where the scroll listeners would not bubble to parent elements and would therefore not be able to listen to events higher in the bubble chain.


rcorral:
LGTM :shipit:

PR #389, opened 2017-01-30
Specify c++ version for Travis build

nLight:
Build is failing because of C++11 issues, this is what I've found https://github.com/travis-ci/travis-ci/issues/4771

Adding these lines seems to fix the build

```yaml
env:
  - CXX=g++-4.8
addons:
  apt:
    sources:
    - ubuntu-toolchain-r-test
    packages:
    - g++-4.8
```

PR #388, opened 2017-01-19
Update stylelint deps

jfurrow:
This PR updates two dependencies. They were causing the build to fail due to recent internal changes in `stylelint`. 

Related:
https://github.com/olegskl/gulp-stylelint/issues/57

PR #390, opened 2017-01-30
Allow react 15.0.0 and 15.4.0

nLight:
~~ℹ️  it also contains code from #389 and #388~~ Merged!

I did `npm run livereload` and all widgets seemed to be working

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,16 @@ node_js:
   - "4.4"
 before_install:
   - npm install -g npm@3.9
+install:
+  - npm install
+  - npm install react@$REACT_VERSION react-addons-css-transition-group@$REACT_VERSION react-addons-test-utils@$REACT_VERSION react-dom@$REACT_VERSION
 env:
-  - CXX=g++-4.8
+  global:
+    - CXX=g++-4.8
+  matrix:
+    - REACT_VERSION=0.14
+    - REACT_VERSION=15.0
+    - REACT_VERSION=15.4
 addons:
   apt:
     sources:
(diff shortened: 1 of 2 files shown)


nLight:
Tests pass locally, interesting

mlunoe:
Here are some specifics about the Semver versions: https://docs.npmjs.com/misc/semver

In general, we should use fixed versions for actual dependencies (normal and dev) except for the ones we require as peer dependencies. Here I guess to goal is to be as relaxed to as possible.

I tend to use `x`s when specifying versions as it is more obvious what is happening than when using `~` or `^`.

I just did a test and it will install the highest version possible when using `||` and `x`s (which is also what I remember, but couldn't find any resources for this, though I didn't do a thorough search for it).

mlunoe:
Also, it seems that we are not ready for react 15.x.x

nLight:
Fair points, thanks for the feedback! I'll continue working on the branch.

nLight:
on the other hand we could just roll with 0.14 and 15.4.x :)

mlunoe:
If we do support multiple versions, we should make sure to support the full range.
I would pretty annoyed as a developer that if a library would skip a version (which could happen to be the one I am using).

nLight:
Hey @mlunoe I believe there was an artifact that broke 15.0 build for you. I added all the versions in travis and it is all green 💚

nLight:
though 15.4.x throws some warnings we will eventually address.

mlunoe:
Great! And thank you for cleaning up the package.json! Looks so much better now!

PR #391, opened 2017-01-31
DCOS-12966: Make virtual list bind its onScroll function correctly

mlunoe:
Closes DCOS-12966

This PR binds the onScroll function to the right context, so we don't hit this error again:


jfurrow:
This works great!~

mlunoe:
We need to create another update of reactjs-components for this to take effect!

PR #393, opened 2017-02-28
DCOS-14141: Fix modal height calculations in IE11

jfurrow:
This PR fixes the modal height calculations for IE11. Here's a gif!

Before:

After:


jfurrow:
I think I found a bug with this, so I'm adding `hold merge` until I can address it.

MatApple:
Nice work @jfurrow ✨  🎉

PR #394, opened 2017-03-01
DCOS-14092: Explicitly set dropdown positioning properties to auto

jfurrow:
This PR explicitly sets positioning values to `auto` rather than relying on their default properties. This fixes a bug in IE11 where the dropdown would be rendered flush with the left edge of the viewport.

Before:

After:


mlunoe:
Looks good dude!

nLight:
giphy-2

PR #398, opened 2017-04-10
DCOS-11781: fix virtual list rendering and update issues

mlunoe:
This PR fixes a problem in VirtualList that would make it not render correctly.
- It was passing down the wrong index to rendered items
- We turned on capturing in the scroll event listener, so it can change the rendered items before any other listener is fired
- The scroll listener is now set to fire on both leading and trailing edge to make sure to minimize flash of unrendered items


wavesoft:
Looks good to me! 👍 ✨

src/Util/Util.js
```diff
@@ -566,25 +566,72 @@ const Util = {
   },
 
   /**
+   * Underscore.js 1.8.3. For more information,
```
nLight: Why don't we use underscore everywhere instead? We copy the function and have different versions flying around. Let's switch to having underscore/throttle as a dependency

src/Util/Util.js
```diff
@@ -566,25 +566,72 @@ const Util = {
...
+   * see http://underscorejs.org/docs/underscore.html#section-82
    * @param {Function} func A callback function to be called
    * @param {Number} wait How long to wait
-   * @returns {Function} A function, that, as long as it continues to be
-   * invoked, will not be triggered. The function will be called
-   * after it stops being called for N milliseconds.
+   * @param {Object} [options]
+   * @param {Object} [options.leading = false] Run function execution on the
```
orlandohohmeier: Is it possible enable/disable both leading and trailing at the same time? It might makes sense to combine these into one flag...

src/Util/__tests__/Util-test.js
```diff
@@ -127,7 +127,10 @@ describe('Util', function () {
       setTimeout(throttled, 200);
       jest.runAllTimers();
 
-      expect(func.mock.calls.length).toBe(2);
+      // The calls should be trhee because #throttle will remember if it
```
orlandohohmeier: 🔤  "three"

src/VirtualList/VirtualList.js
```diff
@@ -20,25 +20,30 @@ class VirtualList extends React.Component {
...
-    // Replace onScroll by debouncing
+    // Replace onScroll by throttling
     if (this.props.scrollDelay > 0) {
       this.onScroll = Util.throttle(
         this.onScroll.bind(this),
-        this.props.scrollDelay
+        this.props.scrollDelay,
+        // Fire on both leading and trailing edge to minize flash of
+        // un-rendered items
+        {leading: true, trailing: true}
```
orlandohohmeier: Wonder if this is the right solution... Seems that we actually want to make sure that the handler is called _n_ times per second when needed. Anyway this is fine for now.

src/VirtualList/VirtualList.js
```diff
@@ -210,6 +215,16 @@ VirtualList.getItems = function (viewTop, viewHeight, listTop, itemHeight,
   return result;
 };
 
+VirtualList.getItemsToRender = function (props, state) {
```
orlandohohmeier: Why is this a _static_ method? Let's turn this one into a public method or an util/private helper.

mlunoe: To make it testable, and following how this was built out in the first place. I wanted to not spend more time on this fix then necessary. If I make that change, I feel we should do it for the whole file.

src/VirtualList/VirtualList.js
mlunoe: What do you mean? Setting down the scrollDelay?
Please elaborate.

src/Util/Util.js
mlunoe: We are using both leading and trailing in our case and I am not going to change the underscore functionality.
I will attempt to include the function from a library instead

src/Util/Util.js
mlunoe: Will give it a try

mlunoe: Done

src/VirtualList/VirtualList.js
mlunoe: Ok, moved everything in, but I really felt that this was not part of the task

src/VirtualList/VirtualList.js
orlandohohmeier: Reading your documentation of the respective options I assumed the functions will be executed twice in a very short time as illustrated below, but the leading/trailing actually only refers to the every first and last call thus we're good.

```
| wait | wait | wait |
 ^    ^ ^    ^ ^    ^
```

mlunoe: Sorry, yeah I guess my documentation didn't convey that message

mlunoe:
@orlandohohmeier how do you feel about the change? Would love to get the ball rolling as this is holding back a high-priority bug fix

src/VirtualList/VirtualList.js
```diff
@@ -7,22 +7,36 @@
...
   constructor() {
     super(...arguments);
 
     this.state = this.getVirtualState(this.props);
 
     // Replace onScroll by throttling
     if (this.props.scrollDelay > 0) {
-      this.onScroll = Util.throttle(
+      this.onScroll = throttle(
         this.onScroll.bind(this),
```
orlandohohmeier: Why do we need to bind `onScroll` again? Shouldn't is be bound to the correct scope by utilizing the bind mixin?

src/VirtualList/VirtualList.js
```diff
@@ -7,22 +7,36 @@
...
+import BindMixin from '../Mixin/BindMixin';
 import Util from '../Util/Util';
 
 let mathMax = Math.max;
 let mathMin = Math.min;
 let mathFloor = Math.floor;
 let mathCeil = Math.ceil;
 
-class VirtualList extends React.Component {
+class VirtualList extends Util.mixin(BindMixin) {
```
orlandohohmeier: Do we really need the `BindMixin`? Would love to slowly reduce the usage of mixins...

src/VirtualList/VirtualList.js
```diff
@@ -7,22 +7,36 @@
 import DOMUtil from '../Util/DOMUtil';
 import React from 'react';
 import ReactDOM from 'react-dom';
+import throttle from 'lodash.throttle';
+
+import BindMixin from '../Mixin/BindMixin';
 import Util from '../Util/Util';
 
 let mathMax = Math.max;
```
orlandohohmeier: Not related to this PR but why are we _"caching"_ those utils? This saves basically nothing and is is esp. silly as we call `getVirtualState` in our on scroll handler which triggers layout/reflow (`getBoundingClientRect`, _`clientHeight`_).

mlunoe: I tested this out and actually had a significant impact when I implemented this in the first place. Let's create a separate task to optimise this

orlandohohmeier:
Code l👀ks good! 🥇 💯

PR #399, opened 2017-04-12
DCOS-14370: Reduce scroll event listener logic in Tooltip

jfurrow:
This PR reduces the scroll event listener logic for the `Tooltip` component. It will now attach its scroll event listeners to the `window` object in the capture phase so that it's the first listener called when a scroll event happens anywhere in the UI.

This means that the consumers of the `Tooltip` doesn't need to worry about which element to attach the event listener.


weblancaster:
LGTM :1st_place_medal:

mlunoe:
It's the other way around. This merges and then we update dcos-ui :)

PR #400, opened 2017-04-13
Fix linting errors on master

mlunoe:
After some changes to VirtualList and Modal we had some linting errors land on master. This PR fixes that.


jfurrow:
Yep, this fixes it!

PR #403, opened 2017-05-08
fix(DCOS-15159): Don't disable left button close button.

weblancaster:
This commit is a continuation of DCOS-15159, per request from design team close/left button shouldn't be disabled.


weblancaster:
Agree @MatApple I believe this could user some refactoring but will leave for another PR.

PR #407, opened 2017-06-21
fix(react-15): exclude prop types from html attributes

nLight:
This PR addresses all the react 15.4.x warnings by filtering the attributes we provide to tags.

### DIFF
--- a/src/Form/FieldCheckbox.js
+++ b/src/Form/FieldCheckbox.js
@@ -87,13 +87,20 @@ class FieldCheckbox extends FieldRadioButton {
       attributes.indicatorClass
     );
 
+    const htmlAttributes = Util.pick(attributes, [
+      "checked",
+      "className",
+      "disabled",
+      "name"
+    ]);
+
     return (
       <label className={labelClasses} key={index}>
         <input
           onChange={this.handleChange.bind(this, eventName, attributes.name)}
           ref={attributes.name}
           type="checkbox"
-          {...attributes}
+          {...htmlAttributes}
         />
         <span className={indicatorClasses} />
         {this.getItemLabel(attributes)}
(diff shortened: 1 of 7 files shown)


nLight:
🙇  It is an anti-pattern, one step at a time. Gotta test it with DC/OS UI thus WIP.

nLight:
Tested, it should be good to review

src/Form/FieldCheckbox.js
```diff
@@ -87,13 +87,19 @@ class FieldCheckbox extends FieldRadioButton {
       attributes.indicatorClass
     );
 
+    const htmlAttributes = Util.pick(attributes, [
```
ahoskins: Should use `exclude()` here to be consistent with the others.

src/List/ListItem.js
```diff
@@ -8,11 +8,12 @@ class ListItem extends React.Component {
     const { props } = this;
     const Tag = props.tag;
 
+    const excludeProps = ["children", "className", "tag", "transition"];
```
ahoskins: In others the excluded fields are inline, should also be inline here!

src/Form/FieldInput.js
```diff
@@ -214,6 +219,7 @@ FieldInput.defaultProps = {
   columnWidth: 12,
   handleEvent() {},
   value: "",
+  startValue: "",
```
ahoskins: This seems to have snuck in, and not actually in scope of PR?

nLight: BUSTED! 👮  
I think that is a residue from an attempt to fix switching uncontrolled to a controlled input. Maybe it even partially works, I'll double check.

src/Form/FieldCheckbox.js
nLight: I'd prefer having `pick` over `exclude`. Allowing what is for sure ok to allow is more strict and less error prone.

nLight:
Hey @ahoskins I addressed your comments! 🙇

PR #411, opened 2017-07-20
DCOS_OSS-920: Filter unknown html attributes

nLight:
This PR introduces a list of allowed HTML attributes and a new utility function to filter all unknown attributes in a sane way.

Bonus track: `scrollContainer` prop has been gone from `Tooltip`, documentation adjusted.


weblancaster:
@nLight why is this necessary? seems like we are over-engineering somewhere specially because React already does that for us.

nLight:
Hey @weblancaster, react indeed does that, but then it complains a lot in the console. Those warnings make it hard to get a real warning when it happens. I want to emphasize that this is more or less a stop gap measure. Meaning that we shouldn't blindly spat all the props, but we should allow and pass only specific stuff.

I would be more than happy not to sanitize `htmlAttributes` but that requires more thorough refactoring.

weblancaster:
@nLight I understand your point and I think it's valid but I also believe when we made the choice to use React we committed to all pros and cons that React comes with, warnings are one of them.

Introducing this custom "check" to the code base will increase our debt and very likely to cause headaches in the future because of the abstraction introduced now.

When balancing the warnings and the code introduced here I vote to keep the warnings because they are not getting in the way yet, it's annoying yes but something acceptable.

bstavroulakis:
@nLight @weblancaster I would sway towards Michael's position. If feels like that we are abstracting too much and moving away from the native elements too much. 

To create a mock HTMLAttribute is a bit future risky as these may be a bit different from browser to browser and from version to version. Also, we cannot keep up with updating this file with all the various attributes as these change from time to time and new ones are added when new native HTML5 elements are introduced.

nLight:
Thanks everyone for the feedback! We've discussed this topic offline with @orlandohohmeier and @Poltergeist and decided not to move on trying to apply a general fix to the issue. Instead we're going to monitor warnings and address issues individually. That being said I extracted the bonus track from this PR #412 and am closing this one.

PR #415, opened 2017-10-11
Add react 16 support

nLight:
Adding react 16 support 🎉 

✅  https://github.com/noeldelgado/react-gemini-scrollbar/pull/36

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,11 +65,10 @@
     "less-color-lighten": "0.0.1",
     "node-libs-browser": "1.0.0",
     "prettier": "1.2.2",
-    "react": "^15.0.0",
+    "react": "^15.0.0 || ^16.0.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -80,9 +79,9 @@
   },
   "peerDependencies": {
     "cnvs": "1.1.x",
-    "react": "^15.0.0",
+    "react": "^15.0.0 || ^16.0.0",
```
weblancaster: Not sure how I feel about having `||` here but well.

nLight: Hey, @weblancaster, what is your concern? We're testing against _all_ versions + I wanted to keep upgrades smooth and have a fallback option. Is it the way I expressed the supported range or having both versions in first place is the concern?

PR #417, opened 2017-11-08
feat(select): create a new select component wrapping the existing dropdown

juliangieseke:
**⚠️  ~~depends on #418~~** 

This component wraps the existing Dropdown, but also adds an input field which
enables us to emit proper onChange events.

Example Usage (can be used for testing):
```
outdated, see comment below
```
This exact example is from the `Create Service > Volumes > Local Volumes` Select, you can replace the existing one after linking the branch from `reactjs-components` and importing it.

Closes DCOS-19313

### DIFF
--- a/docs/src/Select/index.js
+++ b/docs/src/Select/index.js
@@ -1,0 +1,96 @@
+import React from "react";
+
+import CodeBlock from "../components/CodeBlock";
+import ComponentExample from "../components/ComponentExample";
+import ComponentExampleWrapper from "../components/ComponentExampleWrapper";
+import ComponentWrapper from "../components/ComponentWrapper";
+import Select from "../../../src/Select/Select.js";
+import SelectOption from "../../../src/Select/SelectOption.js";
+import PropertiesAPIBlock from "../components/PropertiesAPIBlock";
+
+class SelectExample extends React.Component {
+  render() {
+    return (
+      <ComponentWrapper
+        title="Select"
+        srcURI="https://github.com/mesosphere/reactjs-components/blob/master/src/Select/Select.js"
+      >
+        <p className="lead flush-bottom">
+          To allow a enclosing
+          {" "}
+          <code>form</code>
+          {" "}
+          recognize the change of a
+          {" "}
+          <code>Dropdown</code>
+          , it is necceccary that there is an
+          {" "}
+          <code>input</code>
+          {" "}
+          element which emits a proper
+          {" "}
+          <code>change</code>
+          {" "}
+          event. This is solved by this component.
+        </p>
+        <PropertiesAPIBlock
+          propTypesBlock={"PROPTYPES_BLOCK(src/Select/Select.js)"}
+        />
+        <PropertiesAPIBlock
+          propTypesBlock={"PROPTYPES_BLOCK(src/Select/SelectOption.js)"}
+        />
+        <ComponentExampleWrapper>
+          <ComponentExample>
+            <div className="row row-flex">
+              <div className="column-6">
+                <Select
+                  name={"Select example"}
+                  className="dropdown-select"
+                  onChange={selectedItem => console.log("prop", selectedItem)}
+                  placeholder="Select your Option…"
+                >
+                  <SelectOption value="Foobar" label="Foobar Label">
+                    <strong>Foobar</strong>
+                    <p>Description</p>
+                  </SelectOption>
+                  <SelectOption value="baz" label="baz Label">
+                    <strong>baz</strong>
+                    <p>Description</p>
+                  </SelectOption>
+                </Select>
+              </div>
+            </div>
+          </ComponentExample>
+          <CodeBlock>
+            {`import { Select } from "reactjs-components";
+import React from "react";
+
+class SimpleSelectExample extends React.Component {
+  render() {
+    return (
+      <Select
+        name={"Select example"}
+        className="dropdown-select"
+        onChange={selectedItem => console.log("prop", selectedItem)}
+        placeholder="Select your Option…"
+      >
+        <SelectOption value="Foobar" label="Foobar Label">
+          <strong>Foobar</strong>
+          <p>Description</p>
+        </SelectOption>
+        <SelectOption value="baz" label="baz Label">
+          <strong>baz</strong>
+          <p>Description</p>
+        </SelectOption>
+      </Select>
+    );
+  }
+}`}
+          </CodeBlock>
+        </ComponentExampleWrapper>
+      </ComponentWrapper>
+    );
+  }
+}
+
+module.exports = SelectExample;
--- a/package.json
+++ b/package.json
@@ -16,10 +16,7 @@
     "url": "https://github.com/mesosphere/reactjs-components"
   },
   "license": "Apache-2.0",
-  "files": [
-    "main.js",
-    "lib"
-  ],
+  "files": ["main.js", "lib"],
   "main": "main.js",
   "dependencies": {
     "classnames": "2.2.3",
@@ -85,10 +82,13 @@
   },
   "scripts": {
     "clean": "rm -rf dist docs/dist lib",
-    "dist": "npm run clean && NODE_ENV='production' ./node_modules/.bin/gulp docs:dist",
+    "dist":
+      "npm run clean && NODE_ENV='production' ./node_modules/.bin/gulp docs:dist",
     "predist-src": "rm -rf ./lib",
-    "dist-src": "cp -r ./src ./lib && find ./lib -name '__tests__' -type d -exec rm -r '{}' + && ./node_modules/.bin/babel lib --out-dir lib",
-    "livereload": "NODE_ENV='development' ./node_modules/.bin/gulp docs:livereload",
+    "dist-src":
+      "cp -r ./src ./lib && find ./lib -name '__tests__' -type d -exec rm -r '{}' + && ./node_modules/.bin/babel lib --out-dir lib",
+    "livereload":
+      "NODE_ENV='development' ./node_modules/.bin/gulp docs:livereload",
     "test": "NODE_PATH=node_modules jest --config=jest/config.json --no-cache",
     "release": "npm version -m 'Version %s'",
     "release:major": "npm run release -- major",
@@ -102,7 +102,9 @@
     "lint": "npm run stylelint && npm run eslint",
     "eslint": "eslint --quiet src/**/*.js docs/src/**/*.js",
     "stylelint": "stylelint src/**/*.less  docs/src/**/*.less",
-    "preversion": "npm run clean && npm test && npm run lint && npm run dist-src",
-    "postversion": "git push --tags && npm publish ./ --tag version-$npm_package_version && git push && echo \"Successfully released $npm_package_version\""
+    "preversion":
+      "npm run clean && npm test && npm run lint && npm run dist-src",
+    "postversion":
+      "git push --tags && npm publish ./ --tag version-$npm_package_version && git push && echo \"Successfully released $npm_package_version\""
   }
 }
--- a/src/Select/Select.js
+++ b/src/Select/Select.js
@@ -1,0 +1,100 @@
+import React from "react";
+import PropTypes from "prop-types";
+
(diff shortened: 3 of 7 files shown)


juliangieseke:
new usage example:
```
<Select
  name={`localVolumes.${key}.type`}
  selectWrapperClassName="dropdown-select"
  onItemSelection={selectedItem =>
    console.log("prop", selectedItem)}
  >
  <SelectOption
    id="default"
    title="Select…"
    selectable={false}
    selected={true}
  />
  <SelectOption id="HOST" title="HOST">
    <strong>HOST</strong>
    <span>Recommended</span>
    <p>Description</p>
  </SelectOption>
  <SelectOption id="PERSISTENT" title="PERSISTENT">
    <strong>PERSISTENT</strong>
    <p>Description</p>
  </SelectOption>
</Select>
```

src/Select/Select.js
```diff
@@ -0,0 +1,90 @@
...
+          wrapperClassName={"dropdown"}
+          trigger={InputTrigger}
+          ref={dropdown => (this.dropdown = dropdown)}
+        />
+      </div>
+    );
+  }
+}
+
+Select.defaultProps = {
```
ahoskins: Should also add `Select.propTypes`

src/Select/Select.js
```diff
@@ -0,0 +1,90 @@
...
+  className: "dropdown-select",
+  onChange: undefined,
```
ahoskins: There's no need to set to `undefined` as fallback value, because if not given it will already be `undefined`

src/Select/SelectOption.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+
+class SelectOption extends React.Component {
```
ahoskins: I think it's a bit strange that this component as merely a proxy for passing props which will be used in `Select`. In my opinion the `buildItemsArray` logic from `Select` should be here, since this is the "options" component.

ahoskins:
@juliangieseke You should add an example to `/docs/src/Select` before this merges. This also provides an easy way for reviewers to test this component! 👍

docs/src/Select/index.js
```diff
@@ -0,0 +1,87 @@
...
+                >
+                  <SelectOption
+                    value="default"
+                    label="Select…"
+                    disabled={false}
+                    selected={true}
+                  />
+                  <SelectOption value="HOST" label="HOST">
+                    <strong>HOST</strong>
+                    <span>Recommended</span>
```
ahoskins: I think we want to use the [Badge](http://mesosphere.github.io/cnvs/components/badges/) component here!

src/Select/__tests__/Select-test.js
```diff
@@ -0,0 +1,172 @@
...
+        "dropdown-select input-value"
+      );
+      expect(input.value).toEqual("option-3");
+    });
+    it.skip("TODO: passes correct items array to dropdown", function() {
+      // console.log(this.doc.dropdown.props.items);
+    });
+  });
+  describe("#onChange", function() {
+    /**
```
MatApple: Please make sure not to commit code with commented out tests :)

src/Select/Select.js
juliangieseke: see below ✅

src/Select/__tests__/Select-test.js
```diff
@@ -0,0 +1,108 @@
...
+        this.select,
+        "dropdown-select input-value"
+      );
+      expect(input.value).toEqual("option-3");
+    });
+  });
+  describe("#onChange", function() {
+    /**
+     * doesnt work here, but works in browser...
+    it("fires onChange correctly when dropdown changes", function() {
```
nLight: Let's please either remove this or make it work :)

src/Select/__tests__/Select-test.js
```diff
@@ -0,0 +1,108 @@
...
+          disabled={false}
+          selected={true}
+        >
+          selected
+        </SelectOption>
+        <SelectOption value="option-5" label="option-5-label">
+          <p>One HTML Child</p>
+        </SelectOption>
+        <SelectOption value="option-6" label="option-6-label">
+          <p>Multiple HTML Children</p>
```
nLight: still only one child 👶

src/Select/Select.js
```diff
@@ -0,0 +1,108 @@
...
+          items={this.buildItemsArray()}
+          initialID={this.state.value}
+          onItemSelection={this.handleDropdownChange.bind(this)}
+          buttonClassName={"button dropdown-toggle"}
+          dropdownMenuClassName={"dropdown-menu"}
+          dropdownMenuListClassName={"dropdown-menu-list"}
+          dropdownMenuListItemClassName={"dropdown-menu-list-item"}
+          wrapperClassName={"dropdown"}
+          trigger={<ButtonTrigger placeholder={this.props.placeholder} />}
+          ref={dropdown => (this.dropdown = dropdown)}
```
nLight: Don't see where we're using this ref

src/Select/__tests__/Select-test.js
```diff
@@ -0,0 +1,108 @@
...
+  TestUtils = require("react-dom/test-utils");
+}
+
+var Select = require("../Select.js");
+var SelectOption = require("../SelectOption.js");
+
+describe("Select", function() {
+  beforeEach(function() {
+    this.callback = jasmine.createSpy();
+    this.select = TestUtils.renderIntoDocument(
```
nLight: Will it be possible to use Shallow rendering here? https://reactjs.org/docs/shallow-renderer.html

src/Select/__tests__/Select-test.js
```diff
@@ -0,0 +1,108 @@
...
+        </SelectOption>
+        <SelectOption value="option-6" label="option-6-label">
+          <p>Multiple HTML Children</p>
+        </SelectOption>
+      </Select>
+    );
+  });
+
+  afterEach(function() {
+    Array.prototype.slice.call(document.body.children).forEach(function(node) {
```
nLight: We may not need this with shallow rendering

src/Select/Select.js
```diff
@@ -0,0 +1,108 @@
...
+    this.setState({ value: selectedOption.id });
+
+    this.input.value = selectedOption.id;
+    this.input.dispatchEvent(event);
+  }
+
+  render() {
+    return (
+      <div className={this.props.className}>
+        <input
```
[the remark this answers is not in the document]
juliangieseke: Long story short: no. 
The hack we are using is not compatible with the given Dropdown / InputTrigger setup. Or at least I wasnt able to get it working… 🤔  
I'd suggest not putting more effort into this before we have time to do a proper refactoring.

nLight:
Dropdown PR has been merged. @juliangieseke please rebase/address the outstanding comments. Let's ship it!

src/Select/SelectOption.js
```diff
@@ -0,0 +1,22 @@
...
+import PropTypes from "prop-types";
+
+class SelectOption extends React.Component {
+  render() {
+    return <div>{this.props.children || this.props.label}</div>;
+  }
+}
+SelectOption.defaultProps = {
+  value: null,
+  label: null,
```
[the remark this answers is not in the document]
juliangieseke: added `value` as an additional fallback.

src/Select/__tests__/Select-test.js
juliangieseke: added an evil twin 😈

src/Select/SelectOption.js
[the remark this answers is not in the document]
juliangieseke: And yes, this is basicly an abstract class / interface definition.

src/Select/__tests__/Select-test.js
juliangieseke: not sure why, but when implementing it, the tests never finish... have to investigate this, but I think thats out of scope for this PR.

src/Select/__tests__/Select-test.js
```diff
@@ -0,0 +1,89 @@
+jest.dontMock("../../Mixin/BindMixin");
+jest.dontMock("../../Util/Util");
+jest.dontMock("../../Util/DOMUtil");
+jest.dontMock("../Select");
+jest.dontMock("../SelectOption");
+
+/* eslint-disable no-unused-vars */
+var React = require("react");
+/* eslint-enable no-unused-vars */
+var ShallowRenderer = require("react-test-renderer/shallow");
```
orlandohohmeier: ☠️  `ShallowRenderer` is no longer used — please remove.

src/Select/Select.js
```diff
@@ -0,0 +1,100 @@
...
+Select.defaultProps = {
+  className: "dropdown-select",
+  onChange() {},
+  name: null,
+  placeholder: ""
+};
+
+Select.propTypes = {
+  className: PropTypes.string,
+  onChange: PropTypes.function,
```
orlandohohmeier: ❕  Please make sure that the "PropTypes" version we'er using still provides `function` as the check was renamed to `func`.

docs/src/Select/index.js
```diff
@@ -0,0 +1,98 @@
...
+          <ComponentExample>
+            <div className="row row-flex">
+              <div className="column-6">
+                <Select
+                  name={"Select example"}
+                  className="dropdown-select"
+                  onChange={selectedItem => console.log("prop", selectedItem)}
+                  placeholder="Select your Option…"
+                >
+                  <SelectOption value="HOST" label="Host Label">
```
orlandohohmeier: 🐷  I'd change the example options form DC/OS specific ones to something like "foo"/"bar"

package.json
```diff
@@ -68,6 +68,7 @@
     "react": "^15.0.0 || ^16.0.0",
     "react-dom": "^15.0.0 || ^16.0.0",
     "react-gemini-scrollbar": "2.3.x",
+    "react-test-renderer": "^16.1.1",
```
nLight: Where are we using this renderer?

docs/src/Select/index.js
```diff
@@ -0,0 +1,98 @@
...
+                  <SelectOption value="PERSISTENT" label="Persistent Label">
+                    <strong>Persistent</strong>
+                    <p>Description</p>
+                  </SelectOption>
+                </Select>
+              </div>
+            </div>
+          </ComponentExample>
+          <CodeBlock>
+            {`import {Select} from 'reactjs-components';
```
nLight: Let's use prettified version:
```javascript
import { Select } from "reactjs-components";
import React from "react";

class SimpleSelectExample extends React.Component {
  render() {
    return (
      <Select
        name={"Select example"}
        className="dropdown-select"
        onChange={selectedItem => console.log("prop", selectedItem)}
        placeholder="Select your Option…"
      >
        <SelectOption value="HOST" label="Host Label">
          <strong>Host</strong>
          <span class="badge">Recommended</span>
          <p>Description</p>
        </SelectOption>
        <SelectOption value="PERSISTENT" label="Persistent Label">
          <strong>Persistent</strong>
          <p>Description</p>
        </SelectOption>
      </Select>
    );
  }
}
```

docs/src/Select/index.js
```diff
@@ -0,0 +1,98 @@
...
+    return (
+      <Select
+      name={"Select example"}
+      className="dropdown-select"
+      onChange={selectedItem => console.log("prop", selectedItem)}
+      placeholder="Select your Option…"
+    >
+      <SelectOption value="HOST" label="Host Label">
+        <strong>Host</strong>
+        <span class="badge">Recommended</span>
```
nLight: We should either remove the badge or make it look better
screen shot 2017-11-24 at 17 18 19

nLight: you can use `npm run livereload` to debug docs

src/Select/SelectOption.js
juliangieseke: yeah, it currently is not much more, but it also is a first step in refactoring the dropdown and is resembles the "new" external API we have in mind for this refactoring. thats why wie convert the new children based interface to the old itemsArray one here.

src/Select/Select.js
juliangieseke: yeah, that one is basicly there to show the "default interface props"

PR #418, opened 2017-11-08
Allow custom triggers for Dropdown

nLight:
Sometimes it is convenient to have any tag as a `Dropdown` trigger. `Button` is usually overloaded with styles that is cumbersome to override.


src/Dropdown/Dropdown.js
```diff
@@ -412,14 +416,15 @@ class Dropdown extends Util.mixin(BindMixin) {
...
-          {this.getSelectedHtml(this.getSelectedID(), items)}
-        </button>
+        {React.createElement(
+          triggerTag,
+          {
+            className: props.buttonClassName,
+            onClick: this.handleMenuToggle,
+            disabled: props.disabled
+          },
+          this.getSelectedHtml(this.getSelectedID(), items)
```
juliangieseke: your implementation doesn't allow `triggerTag:"input"` because they arent allowed to have children 😢 
maybe we can change it to receive a ReactComponent instead of a string? 🤔 

IMHO like it is now, it is even worse then being bound to `button`

juliangieseke:
we extended the implementation to receive a component which handles the rendering of the trigger. that way everything should be possible as a trigger and we are also easily able to define any placeholder text directly in the trigger 👍

src/Dropdown/DropdownTriggers.js
```diff
@@ -0,0 +1,67 @@
...
+export class InputTrigger extends ButtonTrigger {
+  constructor() {
+    super(...arguments);
+    this.handleChange = this.handleChange.bind(this);
+  }
+  handleChange(event) {
+    this.props.onChange(event);
+  }
+  getValue(item) {
+    if (item === null) {
```
ahoskins: We usually to double equals for these right? `item == null`

src/Dropdown/DropdownTriggers.js
```diff
@@ -0,0 +1,67 @@
...
+        onClick={this.handleClick}
+        type="button"
+      >
+        {this.getSelectedHtml(selectedItem)}
+      </button>
+    );
+  }
+}
+
+export class InputTrigger extends ButtonTrigger {
```
ahoskins: In my opinion having this extend `ButtonTrigger` only complicates things. The only "super" functionality it uses is `handleClick`.

src/Dropdown/Dropdown.js
```diff
@@ -485,6 +486,7 @@ Dropdown.propTypes = {
   // Transition lengths
   transitionEnterTimeout: PropTypes.number,
   transitionLeaveTimeout: PropTypes.number,
+  // trigger: PropTypes.function,
```
ahoskins: Why is `trigger` in default props, but commented out here in prop types?

src/Dropdown/DropdownTriggers.js
MatApple: +1

src/Dropdown/DropdownTriggers.js
```diff
@@ -0,0 +1,67 @@
...
+  constructor() {
+    super(...arguments);
+    this.handleChange = this.handleChange.bind(this);
+  }
+  handleChange(event) {
+    this.props.onChange(event);
+  }
+  getValue(item) {
+    if (item === null) {
+      return "Select…";
```
MatApple: Please return `null` when no item and make`Select...` the placeholder. We shouldn't have `Select...` as an actual value.

src/Dropdown/__tests__/Dropdown-test.js
```diff
@@ -19,6 +19,41 @@ if (React.version.match(/15.[0-5]/)) {
 var MockDropdownList = require("./fixtures/MockDropdownList");
 var Dropdown = require("../Dropdown.js");
 
+class InputTrigger extends React.Component {
```
MatApple: Why do we declare a new class in this test instead of using an actual component from `./DropdownTriggers`?

ahoskins:
@nLight you should rebase this on latest master, we had #419 merge which fixed some cnvs color issues!

juliangieseke:
addressed comments & rebased :+1:

src/Dropdown/Dropdown.js
```diff
@@ -485,6 +486,7 @@ Dropdown.propTypes = {
   // Transition lengths
   transitionEnterTimeout: PropTypes.number,
   transitionLeaveTimeout: PropTypes.number,
+  trigger: PropTypes.object,
```
nLight: Shall we use `PropTypes.element` instead? It won't work with _any_ object, right?

src/Dropdown/__tests__/Dropdown-test.js
```diff
@@ -270,4 +266,65 @@ describe("Dropdown", function() {
...
+          dropdownMenuListClassName="dropdown-menu-list"
+          items={MockDropdownList}
+          transition={false}
+          wrapperClassName="dropdown"
+        />,
+        document.body
+      );
+      expect(instance.getSelectedItem()).toEqual(null);
+    });
+    it("shows placeholder when initialized with given trigger", function() {
```
nLight: Do these tests belong here? 🤔

src/Dropdown/__tests__/Dropdown-test.js
```diff
@@ -18,6 +18,24 @@ if (React.version.match(/15.[0-5]/)) {
 
 var MockDropdownList = require("./fixtures/MockDropdownList");
 var Dropdown = require("../Dropdown.js");
+var ButtonTrigger = require("../ButtonTrigger.js").default;
+
+class InputTrigger extends ButtonTrigger {
```
nLight: Is there another way we could implement a mock? Looks like the whole thing belongs more to the `<Select />` pr

src/Dropdown/Dropdown.js
```diff
@@ -289,22 +295,18 @@ class Dropdown extends Util.mixin(BindMixin) {
...
-
   getSelectedID() {
     return this.props.persistentID || this.state.selectedID;
   }
 
+  getSelectedItem() {
+    return (
+      this.props.items.find(
+        item => item.id && item.id === this.getSelectedID()
+      ) || null
```
nLight: Will `undefined` be a problem here? `find` will return `undefined` unless it finds something.

nLight:
I can't request changes on my own pr 😅 @juliangieseke please address my comments.

src/Dropdown/__tests__/Dropdown-test.js
juliangieseke: hm, this isnt related to select in any way?!

src/Dropdown/__tests__/Dropdown-test.js
juliangieseke: I could create a `ButtonTrigger-test.js` 🤔

src/Dropdown/__tests__/Dropdown-test.js
juliangieseke: testing input.value is easier then fetching stuff from the DOM, thats why I extended the trigger here.

src/Dropdown/Dropdown.js
```diff
@@ -485,6 +484,7 @@ Dropdown.propTypes = {
   // Transition lengths
   transitionEnterTimeout: PropTypes.number,
   transitionLeaveTimeout: PropTypes.number,
+  trigger: PropTypes.element,
```
ahoskins: And could check for instance of `DropdownTrigger` right here if follow the approach suggested above.

src/Dropdown/ButtonTrigger.js
```diff
@@ -0,0 +1,51 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+
+export default class ButtonTrigger extends Component {
```
ahoskins: Why is the `InputTrigger` and DropdownTriggers.js` file gone?

*Just a suggestion*: what if you make a base class `DropdownTrigger` that is the "interface" a component needs in order to be trigger. And then let `ButtonTrigger`, `InputTrigger`, etc extend this class (and be in the dropdown triggers file?). Right now this PR allows "any tag as a trigger", but I think it should enforce the behavior of trigger components.

src/Dropdown/__tests__/Dropdown-test.js
nLight: Please do this

src/Dropdown/ButtonTrigger.js
juliangieseke: yeah, we discussed this yesterday and thought it will be good enough having a `ButtonTrigger` - but you are right, we should implement an abstract `DropdownTrigger` 👍  
@nLight & @orlandohohmeier are discussing it further, lets see what their conclusion will be.

src/Dropdown/ButtonTrigger.js
```diff
@@ -0,0 +1,51 @@
...
+        {this.getSelectedHtml(selectedItem)}
+      </button>
+    );
+  }
+}
+
+ButtonTrigger.defaultProps = {
+  className: "",
+  disabled: false,
+  onAction() {},
```
orlandohohmeier: 📛  I'd change the name to align the prop with the class name. How about `onTrigger`?

src/Dropdown/ButtonTrigger.js
```diff
@@ -0,0 +1,51 @@
...
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  handleClick(event) {
+    this.props.onAction(event);
+  }
+
+  getSelectedHtml(item) {
+    if (item != null) {
+      return item.selectedHtml || item.html;
```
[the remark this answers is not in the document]
juliangieseke: then this one returns `undefined` which is the expected behaviour.

src/Dropdown/ButtonTrigger.js
```diff
@@ -0,0 +1,51 @@
...
+    super(...arguments);
+
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  handleClick(event) {
+    this.props.onAction(event);
+  }
+
+  getSelectedHtml(item) {
```
orlandohohmeier: 🎨  Do we really need this method to get the respective value? I know that we're having  similar methods in almost all of our components but we should only add them if doing the operation inline, creating a util or "sub" component is not an option.

src/Dropdown/Dropdown.js
```diff
@@ -104,6 +106,10 @@ class Dropdown extends Util.mixin(BindMixin) {
   handleMenuToggle(e) {
     e.stopPropagation();
 
+    if (this.props.disabled) {
```
orlandohohmeier: 🤔  What happens if the dropdown was disabled while it's open?

src/Dropdown/ButtonTrigger.js
```diff
@@ -0,0 +1,51 @@
...
+    if (item != null) {
+      return item.selectedHtml || item.html;
+    }
+
+    return this.props.placeholder;
+  }
+
+  render() {
+    const { className, disabled, selectedItem } = this.props;
```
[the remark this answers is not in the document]
orlandohohmeier: P.S. We even have a `placeholder` prop similar to an input field...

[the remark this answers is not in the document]
juliangieseke: ✅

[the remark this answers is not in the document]
juliangieseke: As the `Dropdown` currently is a `DropdownList` and the `ButtonTrigger` will be renamed to `DropdownListTrigger` it should be okay like it is now.

src/Dropdown/Dropdown.js
nLight: I can ask the same question to the former implementation :trollface:

src/Dropdown/Dropdown.js
juliangieseke: looks like we dont have an easy way checking if `trigger` is an instance of `DropdownTrigger`. /cc @orlandohohmeier

orlandohohmeier: We need to implement a custom validation similar to https://github.com/mesosphere/marathon-ui/commit/8ecd621c1d3ea4f2ec9e516d8823d1b46467bde5#diff-9ed932ecbb81eee77651242eed89601bR250  @nLight could you please file an issue to introduce such a util and it to the foundation backlog?

juliangieseke:
Hm, tried to use snapshot-testing here, but travis (react 15?) dont like it :(
```
FAIL  src/Dropdown/__tests__/DropdownListTrigger-test.js
  ● Test suite failed to run
    TypeError: Cannot read property 'ReactCurrentOwner' of undefined
      
      at node_modules/react-test-renderer/cjs/react-test-renderer.development.js:77:39
      at Object.<anonymous> (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:7676:5)
      at Object.<anonymous> (node_modules/react-test-renderer/index.js:6:20)
      at Object.<anonymous> (src/Dropdown/__tests__/DropdownListTrigger-test.js:4:26)
```

juliangieseke:
I hope everything is fine now 🚢

package.json
```diff
@@ -16,10 +16,7 @@
     "url": "https://github.com/mesosphere/reactjs-components"
   },
   "license": "Apache-2.0",
-  "files": [
```
juliangieseke: thanks prettier 🥇

src/Dropdown/Dropdown.js
nLight: Didn't `PropTypes.instanceOf` work? Or does it check only provided type and fails to check children?

orlandohohmeier: `PropTypes.instanceOf` doesn't work for components as components aren't instance of the perspective class but react element  thus we need to introduce a custom validator.

src/Dropdown/DropdownListTrigger.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import PropTypes from "prop-types";
+import DropdownTrigger from "./DropdownTrigger";
+
+export default class DropdownListTrigger extends DropdownTrigger {
+  handleTrigger(event) {
```
ahoskins: Why not use the `handleTrigger` in the superclass?

juliangieseke: no idea, changed it - good catch!

src/Dropdown/DropdownListTrigger.js
```diff
@@ -0,0 +1,39 @@
...
+    const { className, disabled, selectedItem } = this.props;
+    const html = selectedItem
+      ? selectedItem.selectedHtml || selectedItem.html
+      : this.props.placeholder || null;
+
+    return (
+      <button
+        className={className}
+        disabled={disabled}
+        onClick={this.handleTrigger}
```
ahoskins: Nit: shouldn't this be `super.handleTrigger`?

juliangieseke: Hm, I think not, imagine something like this:
```
class ExtendedDropdownListTrigger extends DropdownListTrigger {
  handleTrigger(event) {
    super.handleTrigger(event);
    SomeExtendedFunctionality.onDropdownTrigger(event)
  }
}
```
=> I'm not aware of calling super from other methods then the overwritten one 🤔

ahoskins: It can be used outside of same method calls...see example of class Rectangle in the link: https://googlechrome.github.io/samples/classes-es6/

juliangieseke: sure it _can_ be used anywhere, but is it useful/necessary? serious question, if you dont explicitly want to prevent further extension, what is the advantage of calling `super.something()` instead of `this.something()`?

ahoskins: It doesn't really matter, you can leave it. Was just a nit-pick.

nLight: I think the advantage is to skip one prototype jump. Instead of forcing Javascript to look at `this` first before going up in the prototype chain we could point to `super` directly @orlandohohmeier

PR #426, opened 2017-11-24
docs(*): bump 0.17

nLight:
Let's hope it works, otherwise we rollback 😅 
⚠️ targeting `gh-pages` branch


nLight:
There's nothing to build, this is why Travis fails.

PR #432, opened 2017-12-13
DCOS-19864: Fix select

juliangieseke:
This PR fixes the issue, that the select wasnt rerender itself when props changed.

### DIFF
--- a/src/Select/Select.js
+++ b/src/Select/Select.js
@@ -5,14 +5,6 @@ import Dropdown from "../Dropdown/Dropdown";
 import DropdownListTrigger from "../Dropdown/DropdownListTrigger";
 
 export default class Select extends React.Component {
-  constructor() {
-    super(...arguments);
-
-    this.state = {
-      value: this.getInitialValue()
-    };
-  }
-
   buildItemsArray() {
     return React.Children.map(this.props.children, child => {
       return {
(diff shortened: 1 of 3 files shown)


juliangieseke:
well…
screen shot 2017-12-13 at 13 00 47

juliangieseke:
Btw: we can get rid of `<SelectOption> if we somehow force all children to have a `value` prop… 🤔

ahoskins:
@juliangieseke I don't know much about why `SelectOption` is there, but I'd say it's out of this PR scope, wouldn't you agree? I think this PR should be the minimal set of work to make Select work, and avoid changing it's API.

src/Select/Select.js
```diff
@@ -64,12 +35,16 @@ export default class Select extends React.Component {
...
             display: "none"
           }}
-          value={this.state.value}
+          value={this.props.value}
           onChange={this.handleInputChange.bind(this)}
         />
         <Dropdown
           items={this.buildItemsArray()}
-          initialID={this.state.value}
+          persistentID={
```
ahoskins: Nice, didn't notice this as a possible situation 👍

src/Select/__tests__/Select-test.js
```diff
@@ -26,6 +26,7 @@ describe("Select", function() {
         className="select-class-1 select-class-2"
         onChange={this.callback}
         name="SelectName"
+        value="option-3"
```
ahoskins: 😄

juliangieseke:
@ahoskins
re `SelectOption`: you're right, i have changed as much as I could argue is prt of the fix, but removing the `SelectOption` would be out of scope.

---

+1 to andrews last comment

PR #438, opened 2018-02-21
fix(VirtualList): prevents render failing on exception

Fabs:
VirtuaList receives as props functions to render its
items when those functions throw, we send react into
havoc and it starts rendering infinite lists. This
fixes this problem by wrapping the functions on try/catch

CLOSES DCOS-21362

Also, trick to test, but you can:
- On your DC/OS UI repo, checkout branch `bug-reproduce-dcos-20507`, commit `b65b398a59c9ac4f784982e77deaddef0345aa50` go to the nodes page and try to filter for Unhealthy, and back. This will set your list into *infinite craziness ™* (from this reproduction PR https://github.com/dcos/dcos-ui/pull/2760)
- link reactjs-components, check out on this branch, and see that the craziness is gone

### DIFF
--- a/src/VirtualList/VirtualList.js
+++ b/src/VirtualList/VirtualList.js
@@ -24,6 +24,14 @@ const METHODS_TO_BIND = [
   "visibleItems"
 ];
 
+const safeCall = f => {
+  try {
+    return f();
+  } catch (e) {
+    console.error(e);
+  }
+};
+
 class VirtualList extends React.Component {
   constructor() {
     super(...arguments);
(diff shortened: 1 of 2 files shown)


src/VirtualList/VirtualList.js
```diff
@@ -237,11 +241,19 @@ class VirtualList extends React.Component {
       Object.keys(VirtualList.propTypes)
     );
 
+    const safeRenderBufferItem = styles => {
```
nLight: I wonder we do we need an ad-hoc function here 🤔 Can't we declare it outside of the render method?

nLight: private package function?

Fabs:
@nLight, done, and also added a test that was not there and did not verify there is content on the list, let's say, if someone, not me, has forgotten the return in `return f()`

Fabs:
@nLight thanks for approving even with your concerns. I think we should talk about get/setters vs lambdas and come up with a style to follow. I will come later and refactor after we talked.

PR #440, opened 2018-08-20
feat(Dropdown): close on window resize

weblancaster:
Enforce dropdow to close on window resize

Closes DCOS-39951

### DIFF
--- a/src/Dropdown/Dropdown.js
+++ b/src/Dropdown/Dropdown.js
@@ -53,6 +53,14 @@ class Dropdown extends Util.mixin(BindMixin) {
     this.container = this.getScrollContainer();
   }
 
+  componentDidUpdate() {
+    if (this.state.isOpen) {
+      global.addEventListener("resize", this.closeDropdown);
+    } else {
+      global.removeEventListener("resize", this.closeDropdown);
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


natmegs:
Looks good! Is there a reason you added the listeners in componentDidMount instead of in componentWillUpdate (if the state is changing to open) like the keydown and scroll listeners? Or do we need the event listener even when the dropdown is closed?

weblancaster:
@natmegs The main reason I went with `componentDidMount` instead `componentWillUpdate` is that react team marked the api `componentWillUpdate` as unsafe and is deprecating, I also don't like the fact that the listeners depend on the state change instead of relying only on upon react llifecycle.

The performance "gains" for this case seems way to low to sacrifice readability.

Let me know if you if strong about this and I can maybe use `componentDidUpdate` to achieve the same result.

nLight:
Oh, someone's going to release `reactjs-components` 🙌 :trollface: 

How would you feel adding listener in `openDropdown` and removing in `closeDropdown`? No binding to the state change needed.

weblancaster:
@nLight I rather go with @natmegs suggestion because in your suggestion the state would only open after the lifecycle finishes and that is not guaranteed.

PR #500, opened 2019-01-25
fix: check all refs before use

nLight:
Turned out we don't check refs before using them. That PR fixes that.

This one also resolves react16 refs :)


DanielMSchmidt:
Thank you for the help 💪 Let's get this through the door

PR #511, opened 2019-02-06
fix(Dropdown): ref was called without current

DanielMSchmidt:
Additionally this adds some safe guards, so we don't fail with missing
properties

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,5 @@ install:
   - ./install_react.sh
 env:
   matrix:
-    - REACT_VERSION=15.6
     - REACT_VERSION=16.7
     - REACT_VERSION=16.8
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,14 @@ node_js:
 
 jobs:
   include:
+    - stage: lint
+      name: "Lint && Dist"
+      script:
+        - npm run lint && npm run dist
+    - stage: test
+      name: "Unit tests"
+      script:
+        - npm run test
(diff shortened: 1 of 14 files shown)


nLight:
What about this ref? https://github.com/mesosphere/reactjs-components/pull/511/files#diff-acc3a01ef84560b7b72c24ff5650dbf0R233

DanielMSchmidt:
That one was too low to reach :D

nLight:
@DanielMSchmidt this PR makes components 16 only, correct? Previously one wouldn't need `.current`

DanielMSchmidt:
Yeah, that is the case. I'll make a second round to see if I missed some somewhere and change the dependency to be react >= 16

nLight:
fails seem to be legit

DanielMSchmidt:
Could you maybe take a look @nLight? From what I understand the references are not instanciated, but I don't really understand why

DanielMSchmidt:
I found this error, I need to check if it's also there on master
```
raven.js:390 Uncaught TypeError: Illegal invocation
    at Select.handleDropdownChange (Select.js?9d8b:82)
    at Dropdown.handleItemClick (Dropdown.js?8e10:160)
```

DanielMSchmidt:
So the select won't do anything, let's get this fixed 💪

nLight:
ok, I think I nailed the last one

DanielMSchmidt:
Nice, so merge and integrate?

PR #515, opened 2019-02-26
Update react transition group

nLight:


TattdCodeMonkey:
This looks good, but should it be a `feat` commit or a `BREAKING CHANGE`?

nLight:
@TattdCodeMonkey it's both:
```
feat: update react-transition-group@2.5.0

BREAKING CHANGE: transitionLeaveTimeout -> transitionExitTimeout
```

The only problem with the change is that _all_ transitions are gone... so something is missing. either css or js setup

src/Dropdown/Dropdown.js
```diff
@@ -395,23 +396,26 @@ class Dropdown extends Util.mixin(BindMixin) {
...
     }
 
-    if (state.renderHidden) {
-      dropdownMenu = (
-        <div key="concealer" className="dropdown-menu-concealer">
-          {dropdownMenu}
-        </div>
-      );
-    }
+    // if (state.renderHidden) {
```
nLight: @TattdCodeMonkey this is used to calc height of the list and then to decided whether it should open up or down. It's broken

src/Dropdown/Dropdown.js
```diff
@@ -395,23 +396,26 @@ class Dropdown extends Util.mixin(BindMixin) {
...
+    // }
 
     if (props.transition) {
       dropdownMenu = (
-        <CSSTransitionGroup
-          transitionName={transitionName}
-          transitionEnterTimeout={props.transitionEnterTimeout}
-          transitionLeaveTimeout={props.transitionLeaveTimeout}
+        <CSSTransition
+          in={state.isOpen && !state.renderHidden}
```
nLight: That was the important bit. I guess we can move `!state.renderHidden` to the line 407 otherwise it transitions while being hidden

src/Modal/ModalContents.js
```diff
@@ -380,18 +381,20 @@ class ModalContents extends Util.mixin(BindMixin) {
...
-        transitionAppear={props.transitionAppear}
-        transitionEnter={props.transitionEnter}
-        transitionLeave={props.transitionLeave}
-        transitionName={props.transitionNameModal}
-        transitionAppearTimeout={props.transitionAppearTimeoutModal}
-        transitionEnterTimeout={props.transitionEnterTimeoutModal}
-        transitionLeaveTimeout={props.transitionLeaveTimeoutModal}
-        component="div"
+      <CSSTransition
+        in={props.open}
```
nLight: modal should be alrighg

nLight:
@TattdCodeMonkey 

TLDR couldn't figure out `List` `ListItem` though
dropdown transitions but fails to detect if it should be opening `up`
modal works
this is how it should be http://mesosphere.github.io/reactjs-components/

nLight:
testing `npm run livereload`

src/Dropdown/Dropdown.js
TattdCodeMonkey: Figured this out. It is because the ref's were broken and not updated for React 16.

nLight: oh well 😅

mperrotti:
About the `List`/`ListItem` transition in:
I see we updated the styles for a list item animating out, but I don't see any changes for animating in.

However, I think the slide in animation is slow and unnecessary, and has a negative impact on perceived performance. I recommend we remove the ListItem transition.

mperrotti:
Getting a few console errors from props being passed from form components to their dom nodes. A good example is `textarea`.
To be fair, this was happening before these changes, so I won't let that block this particular PR.

PR #526, opened 2019-04-08
fix(Table): don't use virtual list for small tables

natmegs:
Only use VirtualList for tables with large data sets.

## Testing

To test in dcos-ui:
- Check out this branch and run `npm run preversion`
- Go to dcos-ui and replace `reactjs-components` version number in package.json with `"file:../path/to/reactjscomponents/repo"`
- Run on soak and look at components that use reactjs-components table (ie Jobs > Detail)

Jobs run history table comparison:
- Before: https://cl.ly/fc46a4c05463
- After: https://cl.ly/34a03ec1ff43

## Tradeoffs

1000 was chosen as a semi arbitrary cutoff for using VirtualList

### DIFF
--- a/src/Table/Table.js
+++ b/src/Table/Table.js
@@ -271,7 +271,7 @@ class Table extends React.Component {
   }
 
   getTBody(columns, data, sortBy, itemHeight) {
-    const buildRowOptions = this.props.buildRowOptions;
+    const { buildRowOptions, virtualizeDataThreshold } = this.props;
     let childToMeasure;
 
     if (itemHeight === 0 && data.length) {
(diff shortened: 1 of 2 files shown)


jaxesn:
I would make sure and double check for styles/classes in dcos-ui usages of this table to make sure tables dont look/act differently depending on which way they are rendered.  Otherwise big 👍

src/Table/Table.js
```diff
@@ -290,6 +290,17 @@ class Table extends React.Component {
       return <tbody>{this.getEmptyRowCell(columns)}</tbody>;
     }
 
+    if (data.length < 1000) {
```
brandonc: I think this should be a prop, because it truly depends on what you're trying to render. For complex tables, 500 may be the right number For simple ones, maybe 5,000. Suggest "virtualizeDataThreshold" with a default of 1,000

brandonc:
@jaxesn This change doesn't affect the markup, just the number of rows that are rendered... Shouldn't affect any callers!

natmegs:
@jaxesn @brandonc We may just have to change this integration test selector when we bump the version in dcos-ui https://github.com/dcos/dcos-ui/blob/6e0edc864794939ab5d421e537c8deaf87cae27f/tests/pages/services/TasksTable-cy.js#L33

nLight:
Legit fails

GeorgiSTodorov:
reactjs-components only wants 1 review, can we merge this?

nLight:
waiting for the semantic release and PR for dcos-ui please 🙏

DanielMSchmidt:
@nLight here ya go: https://github.com/dcos/dcos-ui/pull/3802

PR #551, opened 2019-08-14
Rename React methods

GeorgiSTodorov:
Everything should still work the same as before.

### DIFF
--- a/src/Util/Util.js
+++ b/src/Util/Util.js
@@ -493,9 +493,12 @@ function sortBy(collection, sortProp) {
 
 const lifecycleFunctions = [
   "componentWillMount",
+  "UNSAFE_componentWillMount",
   "componentDidMount",
   "componentWillReceiveProps",
+  "UNSAFE_componentWillReceiveProps",
   "componentWillUpdate",
+  "UNSAFE_componentWillUpdate",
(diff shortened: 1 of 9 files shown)


src/Util/Util.js
```diff
@@ -492,10 +492,10 @@ function sortBy(collection, sortProp) {
  */
 
 const lifecycleFunctions = [
-  "componentWillMount",
+  "UNSAFE_componentWillMount",
   "componentDidMount",
-  "componentWillReceiveProps",
-  "componentWillUpdate",
```
pierrebeitz: if i understand this `lifecycleFunctions`-thing correctly, we might now break stuff as we might override a `componentWillMount` when using mixins. this could make for unexpected behavious (the lifecycle-method of the component not being called).

pierrebeitz: if we wanted to make this code work correctly (and not just "work in the current dcos-ui") we would need to add the missing lifecycle-methods as well: https://reactjs.org/docs/react-component.html

GeorgiSTodorov: > i think we should keep the old method names as well and just add their UNSAFE-variants.

This is a good idea, I will do it.

